PR #9, opened 2021-07-06
Use Imgix as a next/image custom loader

zainfathoni:


zainfathoni:
Closing this PR until we find a proper and scalable way to use `next/image` or any other image optimization alternatives.

PR #19, opened 2021-07-07
Various optimizations

nibraswastaken:
This PR optimizes nearly everything, notable changes:

- Remove unused sources
- Split and move meta tags to respective documents (`_app.tsx` and `_document.tsx`)
- Use `getStaticProps` instead of `getInitialProps`
- Use `cross-fetch` instead of `node-fetch`
- ~~Adding `unstable_runtimeJS`~~
- Enable experimental features
- Use preact in production

PR #33, opened 2021-07-09
Implement sheets fetcher

adityapurwa:
Implement sheets fetcher to fetch [our database in Google Sheets](https://kcov.id/wbw-sheets) into a machine-readable format.

The fetched content will be stored in [data/wbw-sheets.json](data/wbw-sheets.json) with the following structure:

```
[
  {
    id: number; // Id of the sheet
    name: string; // Title of the sheet
    data: [
      // Depending on the structure of the sheet, the first row will become the key of the object in snake_case form
      // Below is an example for "Jkt" Sheet, the first row has "Kebutuhan", "Keterangan", "Lokasi" as its values
       {
          kebutuhan: string;
          keterangan: string;
          lokasi: string;
          ...
          bentuk_verifikasi: string;
       }
    ]
]
```


zainfathoni:
@adityapurwa Could you, please provide the expected JSON structure so any other contibutors can start working on #34?
Thanks! 🙏

adityapurwa:
> @adityapurwa Could you, please provide the expected JSON structure so any other contibutors can start working on #34?
> Thanks! 🙏

I updated the PR description 🙏

etc/fetchers/fetch-docs.js
```diff
@@ -0,0 +1,16 @@
+const cheerio = require("cheerio");
+const fs = require("fs");
+const path = require("path");
+const fetch = require("node-fetch");
```
zainfathoni, line 4: In #19, @nibraswastaken [tried to replace](https://github.com/kawalcovid19/wargabantuwarga.com/pull/19#discussion_r665538380) `node-fetch` with `cross-fetch`.

I think it's a good idea due to [its universal claim](https://www.npmjs.com/package/cross-fetch), so we can use it later in the browser environment when we need it.
Could you please do the replacement as well in this PR? Thanks!

etc/fetchers/fetch-sheets.js
```diff
@@ -0,0 +1,78 @@
...
+    if (item.length !== 0) {
+      return false;
+    }
+  }
+  return true;
+}
+
+module.exports.fetchSheets = async function fetchSheets() {
+  const source = await fetch(
+    "https://docs.google.com/spreadsheets/d/1RIcSiQqPCw-6H55QIYwblIQDPpFQmDNC73ukFa05J7c/htmlview"
```
zainfathoni, line 20:
```suggestion
    "https://kcov.id/wbw-sheets"
```

Please update the URL to use this shortened one instead, so it's easier to change when we need to change it in the future.

zainfathoni (changes requested):
Overall it looks good to me, but I have minor suggestions here, please address them.
Thanks!

etc/fetchers/fetch-sheets.js
adityapurwa: Noted, just to confirm, the kcov.id/wbw-sheets points to the `htmlview` version, correct? Because previously it redirects me to the editable version of the sheets.

adityapurwa: Fixes at ffe2fd1

etc/fetchers/fetch-docs.js
adityapurwa: Fixes at ffe2fd1

etc/fetchers/fetch-sheets.js
zainfathoni: Yes, it is. I have updated the link yesterday. 😅

zainfathoni (approved):
LGTM 💯

PR #42, opened 2021-07-10
Sentry integration

kalwabed:


zainfathoni:
@kalwabed If [the deployment keeps failing](https://app.netlify.com/sites/wargabantuwarga/deploys/60e9ac64e0ad840007ac1cf6) due to `@netlify/plugin-lighthouse` plugin failure, you can decrease the threshold into the actual number temporarily.
It's configured in the [netlify.toml](https://github.com/kawalcovid19/wargabantuwarga.com/blob/main/netlify.toml#L9) file.

.env.local.example
```diff
@@ -0,0 +1,2 @@
+SENTRY_AUTH_TOKEN=
+NEXT_PUBLIC_SENTRY_DSN=
```
zainfathoni, line 2: Could you please update our [GitHub Actions deployment configuration](https://github.com/kawalcovid19/wargabantuwarga.com/blob/main/.github/workflows/deploy.yml) to ensure that the builds done on GitHub Actions are also using the correct environment variables?

You could use this [on push](https://github.com/kawalcovid19/wargabantuwarga.com/commit/5b6df79480e289c5b4d896f5e80305fe63f8c73f#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3) trigger to trigger the build based on pushes to our branch.

netlify.toml
```diff
@@ -6,9 +6,9 @@
   package = "@netlify/plugin-lighthouse"
 
   [plugins.inputs.thresholds]
-    performance = 0.82
+    performance = 0.65
```
zainfathoni, line 9: I think `0.65` is too low, try increasing it into `0.75` instead, since the last build was having the `0.75` score.
```suggestion
    performance = 0.75
```

netlify.toml
```diff
@@ -6,9 +6,9 @@
...
     accessibility = 1
-    best-practices = 0.93
+    best-practices = 0.87
```
zainfathoni, line 11:
```suggestion
    best-practices = 0.93
```
Try reverting this change.
If this Best Practices score degradation is [caused by this `Browser errors were logged to the console` error](https://app.netlify.com/sites/wargabantuwarga/deploys/60ea444bbe51b30008a9ccb6), it's a problem that we have to fix because it's not caused by the Google Docs CMS.
It's highly likely to be caused by our Sentry integration changes in this branch.

Screenshot 2021-07-11 at 15 32 56

pages/index.tsx
```diff
@@ -1,4 +1,6 @@
 import Head from "next/head";
+import NextScript from "next/script";
+
```
zainfathoni, line 3:
```suggestion
```

Please remove this import statement if we're not using it anywhere within this file.

sentry.server.config.js
```diff
@@ -0,0 +1,17 @@
...
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs";
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
+
+Sentry.init({
+  dsn: SENTRY_DSN,
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: 1.0,
```
zainfathoni, line 12: Please put this value into an environment variable as well, because we would like to be able to configure it easily without any code changes, thanks!

zainfathoni (changes requested):
Other than my comments below, please also resolve the merge conflicts with the `main` branch.

Thank you for your contribution, @kalwabed!

.env.local.example
kalwabed: So that means I've to change it like this?
```yml
on:
  push:
    branches: [feature/sentry-integration]
```

zainfathoni: After a second thought, I don't think that's a good idea, because that would eventually push your changes from this branch to production. 😅 
So if you want to do it, we need to create a separate CI script for the sake of testing & building only.
You can refer to [this tests.yml](https://github.com/ariya/dekontaminasi/blob/master/.github/workflows/tests.yml) as an example.

.env.local.example
```diff
@@ -0,0 +1,3 @@
+SENTRY_AUTH_TOKEN=
```
zainfathoni, line 1: Are we using this value anywhere?

I didn't see any usage of this `SENTRY_AUTH_TOKEN` value in this PR. 🤔

pages/_app.tsx
```diff
@@ -30,6 +31,11 @@ export default function App({ Component, pageProps }: AppProps) {
         <meta name="twitter:creator" content="Warga" />
         <link rel="icon" href="/favicon.ico" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <script
+          async
+          src="https://js.sentry-cdn.com/0dda1c67a0744907b58c15b965348e63.min.js"
+          crossOrigin="anonymous"
+        />
```
zainfathoni, line 38: After trying to run `npx @sentry/wizard -i nextjs`, all the changes other than this one are applied.

I didn't see any instruction to include this script, though. Is it even necessary?

zainfathoni (commented):
I need some clarifications about these issues below, please help me to clarify them. Thanks, @kalwabed! 🙏

sentry.server.config.js
```diff
@@ -0,0 +1,15 @@
+// This file configures the initialization of Sentry on the server.
+// The config you add here will be used whenever the server handles a request.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs";
+
+Sentry.init({
+  dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: process.env.NEXT_PUBLIC_SENTRY_TRACES_SAMPLE_RATE,
```
zainfathoni, line 10: The value from the environment variable is a `string`, isn't it? I guess we need to ensure that the value passed here is a proper floating number, e.g. `1.0` instead of `"1.0"`.

zainfathoni, line 10:
```suggestion

const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
const SENTRY_TRACES_SAMPLE_RATE =
  parseFloat(process.env.SENTRY_TRACES_SAMPLE_RATE) || 0.1;

Sentry.init({
  dsn: SENTRY_DSN,
  // Adjust this value in production, or use tracesSampler for greater control
  tracesSampleRate: SENTRY_TRACES_SAMPLE_RATE,
```

The value from the environment variable is a `string`, isn't it? I guess we need to ensure that the value passed here is a proper floating number, e.g. `1.0` instead of `"1.0"`.

.env.local.example
kalwabed: Instead of including `SENTRY_AUTH_TOKEN` in the `sentry.properties` file, we can use environment variables for that. 

Reference: https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#use-a-properties-file

pages/_app.tsx
kalwabed: Yes, I think, referring to https://docs.sentry.io/platforms/javascript/install/lazy-load-sentry/. But somehow it doesn't show any effect. I'll remove it later.

sentry.server.config.js
kalwabed: Oh sure. I'll change it later.

pages/_app.tsx
zainfathoni: I've tried another attempt on another PR #82 and indeed, it [reduced the Performance score significantly](https://app.netlify.com/sites/wargabantuwarga/deploys/60ecf2bfb5ea9a00086e08ac), even without the `script` tag. 😅 

Hence we might need to watch [this video](https://www.youtube.com/watch?v=xEikLxEcNiU) completely and see what's we're doing wrong here. 😅

PR #84, opened 2021-07-15
Clearer setup steps for new contributor

pveyes:
## Description

User should run `yarn fetch-wbw` before running development server


zainfathoni (approved):
Thanks for adding this in! 🙏 

I was enquired about this problem by one of our contributors but I forgot to update the README.md file accordingly. 😅

Issue #68, opened 2021-07-11
Build a mobile-first responsive design for Warga Bantu Warga


baraeb92:
## Overview

Description

 As a user who uses the mobile browser to visit the website I want all important information on the web to be mobile friendly so I can easily be accessing it using my relatively small mobile screen size, per instruction here: [https://material.io/design/layout/responsive-layout-grid.html#columns-gutters-and-margins](https://material.io/design/layout/responsive-layout-grid.html#columns-gutters-and-margins)

## Acceptance Criteria

 “Given a user that mainly browses using a mobile browser, when I visited the WargaBantuWarga website and click on google sheet link then the information is displayed in a mobile-friendly way that at least scored 90 on the accessibility score at web.dev audit” 

See guidelines here: https://web.dev/accessible/ and https://tailwindcss.com/docs/responsive-design

Also, please use the following resources to ensure easier component building and design consistency: 

- https://tailwindui.com/#components
- https://www.heropatterns.com/
- https://headlessui.dev/
- https://heroicons.com/
- https://www.figma.com/community/file/958381589801778722

## Require A QA To Test? 

yes

## Requirement

List all requirements that necessary to be detailed in the ticket

- [x] Create the design using [mobile-first responsive design principles](https://www.uxpin.com/studio/blog/a-hands-on-guide-to-mobile-first-design/)
- [x] Enable text compression and make sure that font is visible during load
- [x] Make sure all button or interactive element size is at least 48x48px

## Current tasks

- [x] #87
- [x] #92 
- [x] #104 
- [x] #122
- [x] #127
- [x] #174
- [x] #241

Issue #38, opened 2021-07-09
Implement an adequate styling for the Netlify CMS-managed home page


zainfathoni:
## Tasks

- [x] #114
- [x] #190

emer7:
Halo, issue ini maksudnya apa ya mas? Permisi kontribusi apabila boleh

_Hello, what is this issue about? Would like to try to contribute if possible_

zainfathoni:
Thanks, @emer7 !

Please create a draft PR to signal the other contributors that you have started working on it, just like what @resir014 did in #91.

FYI, the page to be styled is already implemented in the [/home-page](pages/home-page.tsx) path.

resir014:
Added #114 to the task list since it's somewhat linked to this issue as well.

Issue #31, opened 2021-07-09
Migrate the home page from Google Docs to Netlify CMS


zainfathoni:
## Overview

Based on the Website Improvements #23 discussion, we are planning to migrate our Google Docs-based CMS into a proper Headless CMS. This would provide more granular controls over the HTML structure that we're rendering on the page that allows us to style the website flexibly.

## Implementation Model

- Use [Tina Cloud](https://tina.io/cloud/) as the WYSIWYG headless CMS and store the content into Markdown files within this repository.

## Current Tasks

- [x] ~#37~
- [x] #45
- [x] #109
- [x] #118 
- [x] #38
- [x] #62
- [x] #285
- [x] #39

resir014:
I think we need to expedite these. Unfortunately the HTML content that we pull from gdocs doesn't follow proper HTML hierarchy, as mentioned here:

![image](https://user-images.githubusercontent.com/5663877/126035478-397f9671-b7ee-41b6-85fe-5bc2ee345460.png)

https://twitter.com/ginseladipura/status/1416343727790985221

---

Some people also complained about too many texts in the homepage.

![image](https://user-images.githubusercontent.com/5663877/126035476-fcaf265c-f644-4af2-b20c-1ecd215ebb04.png)

https://twitter.com/deus_xmchn/status/1416291634216673280

zainfathoni:
Agreed. Once we finished with #87, this should be our top priority. 💪

resir014:
A quick fix for the time being is to utilise `@tailwindcss/typography`:

https://tailwindcss-typography.netlify.app/

zainfathoni:
Got it, thanks! Issue created at #109.

mazipan:
So the page under Netlify CMS is on:

- https://www.wargabantuwarga.com/home-page

And the current version is on:

- https://www.wargabantuwarga.com/

resir014:
@mazipan Yes, agree. I think before we fully swap the home page to the Netlify CMS version, we should split the homepage contents first, so we could easily break up each section of the homepage like the planned design on the Figma.

cc: @zainfathoni

zainfathoni:
From the potential changes based on Figma design, do you have any suggestions on how we should split the sections into @resir014?

resir014:
@zainfathoni I already have some mapping in my head right now:

- Welcome Message (single page, rendered on dialog)
- Cara Berkontribusi (single page)
[...]

PR #85, opened 2021-07-15
minor style adjustment for mobile devices

pveyes:
First iteration of #68, not really resolve as this is only minor style adjustment

## Description

Feedback from my wife, when visiting through mobile feels cramped.

My solution was to:
 1. Set the width to 40 rem on larger screen devices, but take full `100vw` on mobile device.
 2. Make necessary changes to both article and header image so header image still take 100% of space, while the article have small padding as it contains mostly text

## Result

Before/After

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -8,6 +8,6 @@
       ]
     }],
     "block-no-empty": null,
-    "unit-whitelist": ["em", "rem", "s"]
+    "unit-whitelist": ["em", "rem", "s", "vw", "vh"]
   }
 }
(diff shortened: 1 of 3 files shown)


pveyes:
should be ready to review, but I'm open to any feedback @zainfathoni

zainfathoni (approved):
Looks good to me 💯 , it's very helpful as a first iteration to maintain this webpage that's still relying on Google Docs as the base styling.

Thanks for the suggestions and contributions, @pveyes!

For the next iteration, in #38 we're planning to style https://www.wargabantuwarga.com/home-page which will be managed using Netlify CMS, so we have more flexibility and freedom in styling the HTML elements.

The whole epic can be seen at #31.

zainfathoni (approved):

pveyes:
Thanks for the pointers, will take a look later when I have more free time 😃

PR #90, opened 2021-07-17
faq sheet fetcher preserve new line in jawaban

### LINKED ISSUE #64: Implement FAQ Page
## Description

Currently, one of the links on our website links to an [FAQ Google Sheets](https://kcov.id/wbw-faq).

We need to provide an easy way to consume its content.

## Implementation Model

The Google Sheets fetcher has been implemented on #33 by @adityapurwa. We can use the same method to fetch this https://kcov.id/wbw-faq Google Sheets.

## Tasks

- [x] #65
- [x] #66

ekamuktia:
Closes #64
https://github.com/kawalcovid19/wargabantuwarga.com/issues/64#issuecomment-881806083

## Description
Modify faq sheet fetcher to preserve new line in jawaban


components/html-parser.tsx
```diff
@@ -0,0 +1,16 @@
...
+export function HtmlParser(props: HtmlParserProps) {
+  const { children } = props;
+  return (
+    <span
+      dangerouslySetInnerHTML={{
+        __html: children.replace(/(?:\r\n|\r|\n)/g, "<br>"),
+      }}
+    ></span>
+  );
+}
```
resir014, line 16: I think we should be able to use [htmr](https://github.com/pveyes/htmr) for this without sacrificing that much performance. Can you confirm Mas @zainfathoni?

etc/fetchers/fetch-faq-sheets.js
```diff
@@ -7,14 +7,21 @@ const FAQ_LINK = "https://kcov.id/wbw-faq";
 module.exports.fetchFaqSheets = async function fetchFaqSheets() {
   const source = await fetch(FAQ_LINK);
   const $ = cheerio.load(await source.text());
+  $.prototype.textln = function () {
+    this.find("br").replaceWith("\n");
+    this.find("*").each(function () {
+      $(this).replaceWith($(this).html());
+    });
+    return this.html();
+  };
```
resir014, line 16: If we decided to go with `htmr` (as mentioned above), these lines are no longer necessary.

resir014 (commented):
Here are my 2 cents. Nothing too major, just thinking of the possibilities of tools that we can use for this.

components/html-parser.tsx
zainfathoni: Yes, we can use `htmr` in our project.

Initialle I intended to use it, but I decided to let it go because I didn't understand clearly its benefits.

Since one of its benefits are clear enough now, let's try to use `htmr` instead of implementing our own tag-escaping logic. 😉

zainfathoni (changes requested):
As @resir014 mentioned in his comment, please update this PR to use `htmr` instead Mbak @ekamuktia. 😁 Thanks! 🙏

ekamuktia:
@zainfathoni @resir014 
sure i'll update it, please note that on fetcher I use .html() instead .text() for jawaban to preserve the br tag

zainfathoni (approved):
Looks good to me 💯 , thanks for your contribution Mbak @ekamuktia !

PR #121, opened 2021-07-17
add page header on FAQ page

resir014:
## Description

Forgot to add the page header on FAQ page.

image


zainfathoni (approved):
Nice catch, thanks, @resir014 !

PR #136, opened 2021-07-18
Power up search functionality

### LINKED ISSUE #96: Implement a dropdown filter based on `Kategori Pertanyaan`

### LINKED ISSUE #111: Use a proper client-side search engine library to power the search functionality in our pages
For better search results, maybe a full-text search can be used?

The potential libraries are:
- https://www.npmjs.com/package/itemsjs (with aggregation to support the filter)
- https://www.npmjs.com/package/minisearch
- https://fusejs.io/

### LINKED ISSUE #112: Implement a dropdown filter based on `Kebutuhan`

### LINKED ISSUE #113: Implement a dropdown filter based on `Lokasi`

### LINKED ISSUE #131: Sort the Contact List by prioritizing the verified contacts over unverified ones

### LINKED ISSUE #133: Sort the Contact List by `Penyedia` alphabetically ascending

ekamuktia:
Closes #111
Closes #112
Closes #113 
Closes #133
Closes #96 
Closes #131 

## Description

Using itemsjs to improve search function with full text search, aggregation for filter, and sort function

## Current Tasks

- [x] #111 
- [x] #112
- [x] #113
- [x] #133
- [x] #96 
- [x] #131 
- [ ] Need to decide rule for data cleanup before applying it


ekamuktia:
@zainfathoni 
since the url for contact is based on order on search result, there's no fixed url for each contact (the problem also exists in current solution)
which one do you think will be better strategy for contact slug:
- add idx for contact data
- generate slug from penyedia name: will better for SEO, constraint: some contacts have empty value on it

zainfathoni:
We can generate the slugs from `penyedia` field, if `penyedia` is empty, we can use `keterangan` field as a fallback value.

Usually one of them is guaranteed to be filled in.

To be safer, if both values don't exist, we can always use the `index` value as the default fallback value for the last resort. 😅

etc/fetchers/fetch-sheets.js
```diff
@@ -6,6 +6,34 @@ const fetch = require("cross-fetch");
...
+  return str.replace(/([a-z])([A-Z])/g, "$1-$2").toLowerCase();
+}
+function getSlug(name, index) {
+  const kebabName = convertToKebabCase(
+    removeSpaces(
+      replaceSpecialCharacterWithSpace(replaceSpacesWithCamelCase(name)),
+    ),
+  );
+  return `${kebabName}-${index}`;
+}
```
zainfathoni, line 36: I think it would be better if we can reuse the functions in `lib/string-utils.ts` and add the new functions that we need into that file as well.

What's stopping us from doing it? Is it due to the language difference (JS vs TS)?

lib/hooks/use-search.ts
```diff
@@ -1,27 +1,91 @@
...
+          order: "asc",
         };
+        return aggregations;
+      },
+      {},
+    );
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const itemsjs = require("itemsjs")(items, configuration);
```
zainfathoni, line 41: Can we import this library dynamically using the Next.js' [dynamic import](https://nextjs.org/docs/advanced-features/dynamic-import) feature here?

pages/provinces/[provinceSlug]/[contactSlug].tsx
```diff
@@ -65,7 +65,7 @@ export const getStaticProps: GetStaticProps = ({ params = {} }) => {
   const province = index ? provinces[index as unknown as number] : null;
   const provinceName = province ? province.name : "";
   const contact =
-    province !== null ? province.data[contactSlug as unknown as number] : null;
+    province !== null ? province.data.find((a) => a.slug == contactSlug) : null;
```
Comment (zainfathoni) [inline pages/provinces/[provinceSlug]/[contactSlug].tsx:68]: Wouldn't the `find` operation here increase the complexity from O(n) to O(n^2) with n = then number of contacts?

We have provinces with hundreds of contacts, but I don't know how much it would slow down our operations here.
Can we use the index instead? Maybe by appending the index to the end of the slug, like what I did with the `provinceSlug`. In fact, that was my reason for appending the index to the `provinceSlug`, so that we can quickly find the province by `provinces[index]` operation with O(n) complexity with n = the number of provinces.

zainfathoni (changes requested):
I have a few concerns about this PR, it might be not directly related to the search functionality, but since this PR is so big, this kind of concern is inevitable. 😅 

Anyway, thanks for the hard work, Mbak @ekamuktia! 🙏

etc/fetchers/fetch-sheets.js
ekamuktia: Yes, I can't simply export `lib/string-utils.ts` because of the language difference. 
Probably work if the configuration is updated or use additional library. But it'll take some time to explore and can affect the other part, so it'll be great if someone can help

lib/hooks/use-search.ts
ekamuktia: `next/dynamic` itself is for Component based, so I can't use it for the library. Maybe it can be used when rendering `components/search-form.tsx` but it won't affect the search library importing.
Still can't figure out the right way to do it.

pages/provinces/[provinceSlug]/[contactSlug].tsx
Comment (ekamuktia) [reply pages/provinces/[provinceSlug]/[contactSlug].tsx]: Because on the province page first load, it takes some time until the search result appear. I want to make the first data is not from search result, but the original data. But since the contact data in json file is not ordered by `penyedia` name, I changed the fetching process to order it first before saving to file. That's why I didn't use provinces[index] here, because the order is based on `penyedia`.

I changed back the ordering to based on the index so provinces[index] can be used and change the contact list ordering on `getStaticProps`. But seems like the performance is still not optimized.

I may be able to continue this after 4 or 5pm, but if someone can review and try optimize it it'll be really helpful

etc/fetchers/fetch-sheets.js
zainfathoni: Got it, I've created a separate issue for it in #144. Thanks!

lib/hooks/use-search.ts
zainfathoni: Okay, I think we can defer lazy-loading the library for now, since it doesn't affect performance significantly.

However, why are you using `require` instead of the `import` statement? If you're using the `import` statement, you wouldn't need to disable the ESLint rule, would you?
What's stopping you from using the `import` statement?

ekamuktia: The library doesn't have type description so I couldn't use `import`, just added .d.ts file and changed `require` to `import`

pages/provinces/[provinceSlug]/[contactSlug].tsx
Comment (zainfathoni) [reply pages/provinces/[provinceSlug]/[contactSlug].tsx]: Let's separate this problem into another issue, so I can merge this PR now.
Could you please help us to create an issue for this particular concern, Mbak @ekamuktia? Thanks! 😁 🙏

zainfathoni (approved):
LGTM 💯 

Thanks for your contribution, Mbak @ekamuktia! 😄 🙏

PR #141, opened 2021-07-18
Remove nth child check for faster performance

adityapurwa:
Closes 

## Description

The query selector that involves `nth-child` check is very expensive, so we remove it to improve the performance. Based on the data at https://github.com/kawalcovid19/wargabantuwarga.com/discussions/139, it could reduce the duration for sheet fetching from 39.98s to 25.41s.

## Current Tasks

- [ ] 
- [ ] 
- [ ]


zainfathoni (approved):
LGTM 💯 

Thanks, @adityapurwa!

PR #142, opened 2021-07-18
remove `min-h-screen` class on error pages

resir014:
## Description

a height of 100vh doesn't work well on child containers, causing a rogue scrollbar to appear.

image

This fixes it.


zainfathoni (approved):
LGTM 💯 

Thanks, @resir014 !

Issue #115, opened 2021-07-17
SEO Improvement for Province & Contact Pages


ekamuktia:
## Description

Improve SEO strategy for province & contact page for better searchability

## Implementation Model

- Use Next SEO to set page meta & open graph data
- Deciding the rule/pattern for each page meta/open graph title & description
For example: 
On province page, the title use "Informasi Faskes & Alkes untuk COVID-19 di Provinsi (Nama Provinsi)" rule
On contact page, with "(Penyedia) - (Keterangan) di (Lokasi)" rule e.g. "Ambulans rumah sakit advent - Ambulans di Bandar Lampung"
Still have no idea for description part. Need suggestion from SEO expert, also need to consider character length for both title & description.
- Change slug for contact page
Currently there's no id on each contact data, and the slug is based on row order. 
I'm not sure how the data is updated, but if a row is removed on the source sheet and a user already shared the contact page url, the url may link to different contact from what the user intended.
Using slug from contact penyedia name can be a better option, as it's more SEO friendly too. But, if this data can be updated, then different strategy should be considered.
- Add title property on the link (esp. internal one)
- Give rel="nofollow" to external links
- Create sitemap.xml

mazipan:
For sitemap, you can try to use [next-sitemap](https://www.npmjs.com/package/next-sitemap).

I think it will be simpler for us

PR #150, opened 2021-07-19
add next-seo + include default SEO config

resir014:
## Description

Adds [next-seo](https://github.com/garmeeh/next-seo) and include all default SEO config to apply to all pages.

This is a starting point for all the remaining SEO-related issues (e.g. #24, #115)

## Current Tasks

- [x] Add `next-seo` plugin
- [x] Include default SEO config
- [x] Cleanup `<head>` tags that are no longer necessary after `next-seo` addition


config.json
```diff
@@ -1,5 +1,6 @@
 {
   "site_name": "Warga Bantu Warga",
   "site_tagline": "Informasi Faskes & Alkes untuk COVID-19",
-  "site_description": "Inisiatif warga untuk berbagi informasi seputar fasilitas kesehatan & alat kesehatan untuk COVID-19. WargaBantuWarga memudahkan publik mencari & mengakses informasi."
+  "site_description": "Inisiatif warga untuk berbagi informasi seputar fasilitas kesehatan & alat kesehatan untuk COVID-19. WargaBantuWarga memudahkan publik mencari & mengakses informasi.",
+  "site_url": "https://wargabantuwarga.com"
```
zainfathoni, line 5: This config file is managed by Netlify CMS as well, so please add it into the `public/admin/config.yml` configuration  to avoid accidental removal.

And the official URL is with `www.` prefix to avoid potential issues caused by the apex domain.
```suggestion
  "site_url": "https://www.wargabantuwarga.com"
```

zainfathoni (changes requested):
I have some minor suggestions related to Netlify domain configuration and Netlify CMS.

Thanks!

resir014:
@zainfathoni Lighthouse SEO score is still taking a hit here, since we're not using the valid domain name for canonical link (we hardcoded the domain instead).

We need to be able to dynamically get the domain from Next.JS so that the `site_url` config is also not necessary.

zainfathoni (approved):
LGTM 💯 

Please create an issue for the canonical problem that you mentioned above @resir014, thanks! 🙏

PR #162, opened 2021-07-19
add CopyButton component

### LINKED ISSUE #154: Create a `Copy to Clipboard` button for each field

### LINKED ISSUE #156: Reduce the clickable area of the contact item so users can select & copy the `Kontak` & `Alamat` from the list view

### LINKED ISSUE #157: Provide a `Copy to Clipboard` button for `Kontak` & `Alamat` fields in the list view

ekafyi:
Closes #154, closes #157, _possibly_ closes #156

## Description

add `CopyButton` component to copy text content to use

## Current Tasks

- [x] create functioning CopyButton component
- [x] enhance style — use `clipboard-copy` icon from heroicons.com
- [x] implement in ContactList and ContactDetails views


resir014:
@ekafyi Looks like the click event from the parent `<a>` bleeds over to the copy button. Can we put the copy button in the details page instead?

image

ekafyi:
@resir014 There are two issues for this:
- #154 is for the one in your screenshot, this is still in progress
- #157 for the "list view" (does this refer to the Contact List page or did i misunderstand this?)

Next push will be for the one in your screenshot, once i replace the text with HeroIcon component.

resir014:
> for the "list view" (does this refer to the Contact List page or did i misunderstand this?)

Yeah, I think that's for the list view. Then we might need to fix the HTML hierarchy so that the links + buttons don't conflict w/ each other (something similar to [this](https://heydon.github.io/Inclusive-Components/cards-pseudo-content-author-link/) perhaps?)

zainfathoni:
@ekafyi, beware that it might be conflicting with my changes in #151 because we started rendering some of the fields in HTML instead of plaintext.

ekafyi:
> > for the "list view" (does this refer to the Contact List page or did i misunderstand this?)
> 
> Yeah, I think that's for the list view. Then we might need to fix the HTML hierarchy so that the links + buttons don't conflict w/ each other (something similar to [this](https://heydon.github.io/Inclusive-Components/cards-pseudo-content-author-link/) perhaps?)

exactly, was planning to do this once i've got the CopyButton UI done.

resir014:
Also of note: this issue might be related as well.

#156

ekafyi:
what do yall think about the refactored link and copy button UI @resir014 @zainfathoni ? https://deploy-preview-162--wargabantuwarga.netlify.app/provinces/dkijakarta-1

this is still an in-progress + buggy version (copy entire unstripped HTML string), just want to show the link behaviour.

also on an unrelated note, I use `ClipboardCopyIcon` from heroicons, but is that intuitive enough as a "copy" button? 🤔 Maybe we should use the `DuplicateIcon` instead?

components/copy-button.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  return (
+    <button
+      aria-label="salin"
+      className="relative z-10 w-8 h-6 ml-2 bg-gray-100 text-gray-600 hover:bg-blue-600 hover:text-white flex-none flex items-center justify-center rounded"
+      onClick={handleCopy}
+      type="button"
+    >
+      {isCopied ? <span>✓</span> : <ClipboardCopyIcon className="h-4 w-4" />}
```
zainfathoni, line 41:
```suggestion
      {isCopied ? <span>Tersalin ✓</span> : <span>Salin <ClipboardCopyIcon className="h-4 w-4" /></span>}
```

I think adding a word here helps in:

1. Making the actions more obvious to the users
2. Increasing the tap area so it's easier to tap using a thumb on mobile devices

zainfathoni (changes requested):
Overall, it looks good to me.
But I think we can experiment by adding an explanation word beside the icon.

components/copy-button.tsx
ekafyi: Sure good idea @zainfathoni , but I switched the icon-text order in the last push. Now lets wait for the deploy... 😅

zainfathoni (approved):
LGTM 💯 

Thanks for working on it, Mbak @ekafyi!

PR #168, opened 2021-07-19
Change default ordering in province page to `Terverifikasi`

### LINKED ISSUE #164: Change the default sort on the first load of `ProvincePage` from `Nama` into `Terverifikasi`
Currently, we have two sorting rules (#133 & #131) with #133 as the default.

We need to change the default to be #131 instead, because that was the expected behaviour when I created those tickets. 😅

ekamuktia:
Closes #164 

## Current Tasks

- [ ] 
- [ ] 
- [ ]


zainfathoni (approved):
LGTM 💯 

Thanks, Mbak @ekamuktia!

PR #172, opened 2021-07-19
refactor(fetcher): Migrate to ts-node

### LINKED ISSUE #144: Migrate `/fetchers/*.js` files to TypeScript so we can share codes with the `/lib/*.ts` ones
Yes, I can't simply export `lib/string-utils.ts` because of the language difference. 
Probably work if the configuration is updated or use additional library. But it'll take some time to explore and can affect the other part, so it'll be great if someone can help

__Originally posted by @ekamuktia in https://github.com/kawalcovid19/wargabantuwarga.com/pull/136#discussion_r671910801__

## Tasks

- [ ] Use `ts-node` instead of `node` to execute fetcher scripts.
- [ ] Add JSDoc types to all helper functions in fetcher files.
- [ ] Migrate all helper functions in fetcher files to `/lib/*.ts`.

Namchee:
Closes #144 

## Description

Migrates all fetching-related scripts to TypeScript

## Current Tasks

- [x] Use `ts-node` instead of `node` to execute fetcher scripts.
- [x] Add JSDoc types to all helper functions in fetcher files.
- [x] Migrate all helper functions in fetcher files to `/lib/*.ts`.


etc/fetchers/fetch-sheets.ts
```diff
@@ -1,57 +1,24 @@
...
-    if (item.length !== 0) {
-      return false;
-    }
-  }
-  return true;
-}
+import cheerio from "cheerio";
+import fetch from "cross-fetch";
+import fs from "fs";
+import path from "path";
```
mazipan, line 11: Import from 3rd parties are better at first line.
So, please move the import from `utils` after import from 3rd parties.

mazipan (changes requested):

etc/fetchers/fetch-wbw.ts
```diff
@@ -0,0 +1,42 @@
...
+  const spinner = ora(`${chalk.yellowBright("Fetching all data...")}`).start();
+
+  fetchFaqSheets()
+    .then(() => {
+      const end = `${toSecond(process.hrtime(start))} seconds`;
+      spinner.succeed(`Fetching FAQ done in ${chalk.greenBright(end)}`);
+      spinner.start(`${chalk.yellowBright("Fetching next data...")}`);
+    })
+    .catch((err) => {
+      console.error(err);
```
mazipan, line 21: Can we add `chalk.red(err)` here?

etc/fetchers/fetch-sheets.ts
Namchee: Doing that will cause conflicts with [simple sort rules](https://github.com/lydell/eslint-plugin-simple-import-sort). Should we ignore it?

etc/fetchers/fetch-wbw.ts
Namchee: Solved by fee6f4f698fb9a9153742613218f0c18285869f8

etc/fetchers/fetch-sheets.ts
mazipan: See on the example:

https://github.com/lydell/eslint-plugin-simple-import-sort#example

The `react` always coming first, before the internal code.

mazipan: The other tips for node.js, you need to add a newline to seperate the native node.js API such as `fs` and `path` with any other libraries

mazipan:
```ts
import fs from "fs";
import path from "path";

import cheerio from "cheerio";
import fetch from "cross-fetch";

// ...internal import
```

Namchee: It seems that the code from [here](https://github.com/kawalcovid19/wargabantuwarga.com/pull/172#discussion_r672801955) cannot be commited except if we skipped the pre-commit hooks.

The problem is caused by the pre-commit hooks which executes `eslint --cache --fix` which will automatically formats `fetch-sheets.ts` to the current code. Should we temporarily bypass the pre-commit hook for this commit?

mazipan: Just ignore it for now.

Maybe need to think to fix the linter.

mazipan (approved):
All good with a note about the import sorting.
Maybe need to fix the linter 🤔

resir014 (approved):

adityapurwa (approved):

zainfathoni (approved):

PR #196, opened 2021-07-20
add `package-lock.json` to gitignore

resir014:
## Description

Some new contributors keep using `npm` even though the project itself uses `yarn`. This PR add `package-lock.json` to the gitignore file so that people won't accidentally commit npm's lockfile to the project.

In the future, we should explicitly mention in the README that **we use yarn, not npm**.


zainfathoni (approved):
Nice idea 💯

Thanks, @resir014!

PR #207, opened 2021-07-20
fix running `fetch-wbw` script on Windows

resir014:
## Description

On Windows, running the `fetch-wbw` script results in a syntax error since `TS_NODE_COMPILER_OPTIONS='{"module":"commonjs"}'` is not a valid syntax to set environment variables on Windows.

image

Therefore, we use `cross-env` to provide a cross-platform solution in setting environment variables.


zainfathoni (commented):
This is one of the perks of having a maintainer using Windows as his development machine. 😎

PR #217, opened 2021-07-20
add ts-node compiler options in `tsconfig.json`

resir014:
## Description

Turns out adding a `ts-node` key to `tsconfig.json` will make `ts-node` load that specific config. We can remove the environment variable setup this way.

Reference: https://github.com/TypeStrong/ts-node/issues/606#issuecomment-670055983

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "scripts": {
     "build": "next build",
     "dev": "next",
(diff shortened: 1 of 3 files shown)


zainfathoni (approved):
Nice findings, @resir014! 💯

PR #218, opened 2021-07-20
fix: faq page filter

### LINKED ISSUE #203: Search Issue in FAQ page
## Description

There's a minor bug in search in FAQ page. 

1. Type "isoman" then Search
2. Choose filter "PCR"
3. Change the keyword to "vaksin" then Search
4. No results found, the filter lists change to "Semua" and "Vaksinasi" and the active filter changes to "Vaksinasi"

## Expected Behaviour

In step 4, there should be results displayed

## Additional Information

I think in step 3, the filter submitted to the form is still "PCR", thus any keyword changes after that still searches for "PCR" filter even if the active filter indicator already changes to "Vaksinasi" (or "Semua" in the case of no filters found for the keyword). To solve the bug, the user can reload the page then try again, click the reset button, or manually change the filter in step 4 to trigger the filter change hook.

This might be a rather edge case, but I imagine there will be a few people who are searching for multiple unrelated keywords and might be confused of no results during their searching process, especially once the FAQ database grows larger.

mazipan:
Closes #203

## Description

Fix bug on FAQ filter that showing a wrong data when we search any keywords

## Changes

- [x] Add new prop `checkDocSize: boolean` on SearchForm
- [x] Set to false on `faq` page.

## Screenshots

Before

Screen Shot 2021-07-20 at 17 49 36

After

Screen Shot 2021-07-20 at 17 49 49


mazipan:
Help @ekamuktia to check

zainfathoni (approved):
LGTM 💯 

Thanks, Mas @mazipan for working on it.
Mbak @ekamuktia, please help to review these changes as well, thanks! 🙏

components/search-form.tsx
```diff
@@ -161,17 +164,29 @@ export function SearchForm({
...
                           <option
                             key={`option-${key}-${bIdx + 1}`}
                             value={bucket.key}
                           >
                             {bucket.key}
                           </option>
-                        )
-                      );
+                        );
+                      }
```
ekamuktia, line 188: For this part, buckets that have `doc_count` 0 is still printed even though the checkDocSize is true because it goes to the 2nd return. So now, all the filters are always shown.

If the intention for FAQ page is to always show all the categories, another way to do it is to generate `filterItems` value from groupped `faq` instead of using value generated from `useSearch`. 
This way, the filter category order can be in the same order as how the FAQ items are shown which probably is more natural for FAQ page.
image

Sorry, it's a closed pull request but I just have time to check it.

ekamuktia (commented):
Could you check on this part? 
Sorry, it's a closed pull request but I just have time to check it.

components/search-form.tsx
zainfathoni: Could you please report another issue for this Mbak @ekamuktia? So we can do a rework later. Thanks! 🙏

ekamuktia: Okay~!

mazipan: https://github.com/kawalcovid19/wargabantuwarga.com/pull/229

mazipan: For option no 2, I think it will need more time to be done, since it need to refactor the way we select the selected filter.

Issue #225, opened 2021-07-20
Dev Experience: Absolute imports and module path alias


faizrktm:
## Description

Adding/configure absolute imports and module path alias for readability and easier to write for deep and scattered reference.

## Implementation Model

Using `~/` as a leading character for the local module as a distinguishment between the 3rd party lib.

example:

from
import { LayoutRoot } from "../../../components/layout/layout-root";

to
import { LayoutRoot } from "~/components/layout/layout-root";

## Tasks

- [ ] Initial configuration in tsconfig.json 
- [ ] Update current modules path reference

resir014:
I think let's use `~/` instead (e.g. `~/components`) - or have the option to choose between the two.

faizrktm:
sure, let's use that format, I think its more clear to distinguish and still got the benefit from the absolute import. Thanks @resir014

PR #226, opened 2021-07-20
Absolute imports and module path alias

faizrktm:
Closes https://github.com/kawalcovid19/wargabantuwarga.com/issues/225

## Description

Configure tsconfig.json for module path alias with leading `~` from base URL (`./`).

example: `import { PageContent } from "~/components/layout/page-content";`

Nextjs and eslint use tsconfig, might need to update other utilities such as test runner later when it exists to support the path alias.

ref: https://nextjs.org/docs/advanced-features/module-path-aliases

## Current Tasks

- [x] configure tsconfig 
- [x] update current path references

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,15 @@
   "rules": {
     "@babel/new-cap": "off",
     "@typescript-eslint/no-explicit-any": "warn",
-    "import/extensions": ["warn", "never", { "css": "ignorePackages" }],
+    "import/extensions": [
+      "warn",
+      "never",
+      {
+        "css": "ignorePackages",
+        "json": "ignorePackages",
+        "md": "ignorePackages"
+      }
+    ],
     "import/newline-after-import": "warn",
     "import/no-extraneous-dependencies": "warn",
     "import/order": "off",
--- a/components/contact-details.tsx
+++ b/components/contact-details.tsx
@@ -1,7 +1,8 @@
-import { CopyButton } from "@components/copy-button";
-import { anchorTransformer } from "@lib/htmr-transformers";
-import { Contact } from "@lib/provinces";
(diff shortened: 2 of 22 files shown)


resir014 (changes requested):
Can we use `~/` instead? e.g. `~/components`.

mazipan:
Why I personally also support `~` instead of `@` as an alias is because of current situation where we face many libraries published under the scope which is using `@`.

So it become weird, if we can not differentiate our internal code vs the code from 3rd parties.

faizrktm:
yes agree, I also found the format clash with the current installed 3rd party lib `@heroicons`. It's updated! thanks, folks.

btw should we define for each module like `"~/components/*": "./components/*";` to handle possible folder structure changes in the future so we can migrate slowly and safely by only change alias instead of actual import statement.

or we can just `"~/*": "./*"` and agree structure will never change or be willing to change all import statements that changed.

mazipan:
If we learn from the framework approach like Vue CLI, it will come with the default alias ONLY for the root.

So instead of define many alias like

```json
{
  "~/*": "./*",
  "~components/*": "./components/*",
  "~pages/*": "./pages/*",
  "~assets/*": "./assets/*"
}
```

They choose to only define one alias for the root.

```json
{
  "~/*": "./*"
}
```

mazipan (approved):
Looks good for me ✅

faizrktm:
```
Also, you don't need to change all the import on this PR.
Better to do incremental changes.
Let the author update their implementation step by step.
```

well, it's done 😆 since no build is failing, I think it's safe for current changes to be merged ya?

resir014 (approved):
Yep, all good from me too!

faizrktm:
Thanks gengs!

zainfathoni:
@all-contributors please add @faizrktm for code and tools.

PR #227, opened 2021-07-20
feat: seo things

### LINKED ISSUE #115: SEO Improvement for Province & Contact Pages

mazipan:
Closes #115 

## Description

Add some SEO related things

## Current Tasks

- [x] Dynamic robots.txt
- [x] Increase sitemap size to 1000
- [x] Add a basic meta title on Province List Page, Province Detail Page, Contact Detail Page
- [x] Add breadcrumb jsonld using a manual markup -- I decide using a manual way (it will be easier using `next-seo`), for learning purpose, so other developer will know how to generate JSONLD from zero
- [x] Add theme-color same as our header HEX


ekamuktia (dismissed):
image
image
"Warga Bantu Warga" part takes too much space. How about moving it to the back instead?

mazipan:
It's on the default SEO set by @resir014 before.
The common use, the identity sitename is on the right side.
After the main title.

So ya, I agree to move it out to the back.

zainfathoni (approved):
LGTM :100:

Thanks for the initiative, Mas @mazipan! :pray:

PR #229, opened 2021-07-20
fix: conditional on filter

mazipan:
Solve comment on #218


ekamuktia (approved):
LGTM

The similar search issue still happens on `Province` page because of the search rule, but I think we need to discuss the rule first for that issue.

PR #239, opened 2021-07-21
design system - button

### LINKED ISSUE #212: Design system - Buttons
## Description

Expand our current button components for more customisability.

## Implementation Model

Buttons are currently taken from [Tailwind UI](https://tailwindui.com/components/application-ui/elements/buttons).

## Tasks

- [ ] Add White Button variant
- [ ] Expand each button's size options (xs, sm, md, lg, xl)
- [ ] Allow for rounded buttons
- [ ] Allow option to add icons in buttons (from Heroicons)

resir014:
Closes #212

## Description

Adds various features to our existing button components.

Buttons are based on Tailwind UI: https://tailwindui.com/components/application-ui/elements/buttons

## Current Tasks

- [x] Add White Button variant
- [x] Expand each button's size options (xs, sm, md, lg, xl)
- [x] Allow for rounded buttons
- [x] Allow option to add icons in buttons (from Heroicons)
- [x] Refactor button helper functions
- [x] Apply to existing button components
- [x] Add button loading states


resir014:
@hanihusam You might also want to help me review this 😅

components/ui/button/types.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+interface ButtonBaseProps {
+  /** Make the button expand the width of the container. */
+  block?: boolean;
+  /** Is the button rounded? */
+  rounded?: boolean;
+  /** The size of the button. Available sizes: `"xs" | "sm" | "md" | "lg" | "xl"` */
+  size?: ButtonSizes;
+  /** The icon component (from Heroicons) to use in the button. `icon={PlusIcon}` */
+  icon?: React.ComponentType<React.SVGProps<SVGSVGElement>>;
```
mazipan, line 11: do we have ghost button, a button for secondary action?

resir014: We do have a `SecondaryButton` although it sill has a light background - we don't have `GhostButton` just yet, but I think I'll replace `WhiteButton` with it.

mazipan: Ah, do `WhiteButton` means a button that look like a Link ya?
That's different case.

hanihusam:
Do we need loading props to handle a loading which is triggered by button clicked? So there will be a spinner icon and the button will be disabled.

Here is the [reference](https://ant.design/components/button/#components-button-demo-loading) I see.

components/ui/button/types.tsx
```diff
@@ -0,0 +1,15 @@
+export type ButtonSizes = "lg" | "md" | "sm" | "xl" | "xs";
+
+interface ButtonBaseProps {
```
hanihusam, line 3: Do we need loading props to handle a loading which is triggered by button clicked? So there will be a spinner icon and the button will be disabled.

Here is the [reference](https://ant.design/components/button/#components-button-demo-loading) I see.

components/ui/button/types.tsx
resir014: @mazipan No, it's [this one](https://tailwindui.com/components/application-ui/elements/buttons#component-109c4104d58d9fedfa8650dbe24c1ae8):

image

components/ui/button/types.tsx
resir014: Yes, we should. I'll add it to the todo.

components/ui/button/types.tsx
mazipan: Ah this is okay 👍

mazipan (approved):
Looks good to me

mazipan:
Add the Tailwind playground to the description will help @resir014

resir014:
@mazipan I added it on the PR description. If you're talking about the JSDoc comments on each components, it's not really too much. We can add JSDoc comments for each component for all I know.

mazipan:
I mean, something like this https://play.tailwindcss.com/m4d4vqyICn

resir014:
@mazipan ah, I see. I'll see what I can do.

resir014:
@mazipan I created an example playground for the `PrimaryButton` here:

https://play.tailwindcss.com/PldANNK8UZ

All in all, it's possible, however if we're going to update its styles, it will be a pain to update them, since I'll have to re-generate them and publish them here.

We can solve this problem with Storybook, however I think it's a bit too overkill at this early stage.

mazipan (approved):
I think all good to me.

At least almost all our current usecases already covered.
We can add more in the future, but let's keep it simple.

You can merge when you think you're ready

PR #278, opened 2021-07-21
Hide progress bar on shallow routing

### LINKED ISSUE #277: Hide progress bar on shallow routing
## Description

On pages with search feature, the url is updated after keyword/filter/sort is changed using shallow routing. 
The progress bar is quite distracting in these pages.

## Expected Behaviour

Progress bar should not show on shallow routing event, especially because loading state for search will be managed separately.

ekamuktia:
Closes #277

## Description

Check if the router event is shallow event, if yes then do not start nprogress

## Current Tasks

- [x] Update router event to start nprogress only for non shallow event
- [x] Test on province, contact, and faq page by changing keyword, filter, and sort
- [x] Add screen recording from testing

Province List
ezgif com-gif-maker

Contact List
ezgif com-gif-maker (1)

FAQ
ezgif com-gif-maker (2)


pages/_app.tsx
```diff
@@ -22,8 +22,10 @@ const meta = {
 
 export default function App({ Component, pageProps, router }: AppProps) {
   useEffect(() => {
-    const handleStart = () => {
-      NProgress.start();
+    const handleStart = (_: string, { shallow }: any) => {
```
resir014, line 25:
```suggestion
    const handleStart = (_: string, { shallow }: { shallow: boolean }) => {
```

Make sure to not use `any`.

resir014 (changes requested):

pages/_app.tsx
ekamuktia: Thanks, updated!

resir014 (approved):
🚢

mazipan (approved):
Nice ✅

A simple touch

PR #280, opened 2021-07-21
Add scroll to top functionality

### LINKED ISSUE #254: Scroll to top action for long enough scroll content
## Description

image

## Implementation Model

For better UX, it might be better to help user back to the top when he far away scrolling down

## Tasks
- [ ] add button component to bring user to the top of page

kevinmingtarja:
Closes #254

## Description

Created a new ScrollArrow component. Open to suggestions on where it is best to place this button.

Mobile view gif:

Desktop view:
Screen Shot 2021-07-22 at 16 49 36

## Current Tasks

- [X] add button component to bring user to the top of page
- [X] add the component to the pages that needs it


components/ui/scroll-arrow.tsx
```diff
@@ -0,0 +1,37 @@
...
+  };
+
+  return (
+    <ArrowCircleUpIcon
+      className={clsx(
+        "fixed cursor-pointer bottom-20 right-10 h-16",
+        showScroll ? "flex" : "hidden",
+      )}
+      onClick={scrollTop}
+    />
```
resir014, line 35: Please wrap this inside a `<button>`. Don't use `onClick` on non-interactable components.

components/ui/scroll-arrow.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { useCallback, useEffect, useState } from "react";
+
+import { ArrowCircleUpIcon } from "@heroicons/react/solid";
+import clsx from "clsx";
+
+export const ScrollArrow = () => {
+  const [showScroll, setShowScroll] = useState(false);
+
+  const checkScrollTop = useCallback(() => {
+    const heightThreshold = 400;
```
resir014, line 10: Nitpick: We could make this changeable with a prop, I think.

components/ui/scroll-arrow.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { useCallback, useEffect, useState } from "react";
```
resir014, line 1:
```suggestion
import { useCallback, useEffect, useState } from "react";
```

Nitpick: We no longer need to import `React` if we already imported another function.

resir014 (changes requested):
My main issue only relates to a11y right now. Other stuff are just nitpicks.

andriawan:
Do we need seperate issue for implementing a11y @resir014 ? It think it needs more attention related to a11y

resir014:
@andriawan For now we can just focus on making this a proper button. Any further a11y improvements can be made on a different issue.

@kevinmingtarja You can use the circular button example as provided in [Tailwind UI](https://tailwindui.com/components/application-ui/elements/buttons#component-6702d62917fd70644340f4008b126943).

components/ui/scroll-arrow.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <button onClick={scrollTop}>
+      <ArrowCircleUpIcon
+        className={clsx(
+          "fixed text-blue-600 cursor-pointer bottom-20 right-10 h-16 opacity-60 hover:opacity-100 transition duration-300 ease-in-out z-50",
+          showScroll ? "flex" : "hidden",
+        )}
+      />
+    </button>
```
resir014, line 38:
```suggestion
    <button
      className={clsx(
        showScroll ? "inline-flex" : "hidden",
        "items-center p-2 border border-transparent rounded-full shadow-sm text-white bg-blue-600 hover:bg-blue-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-blue-500"
      )}
      onClick={scrollTop}
    >
      <ArrowUpIcon
        className={"h-5 w-5"}
      />
    </button>
```

components/ui/scroll-arrow.tsx
```diff
@@ -0,0 +1,40 @@
+import { useCallback, useEffect, useState } from "react";
+
+import { ArrowCircleUpIcon } from "@heroicons/react/solid";
```
resir014, line 3:
```suggestion
import { ArrowUpIcon } from "@heroicons/react/outline";
```

resir014 (changes requested):
Small design pass for the button. Used the [Tailwind UI circular button](https://tailwindui.com/components/application-ui/elements/buttons#component-6702d62917fd70644340f4008b126943) for this.

ekamuktia:
How about adding the component on `LayoutRoot` instead? This way you don't need to add it on  every page
Is there any page that doesn't need scroll to top button?

andriawan:
good idea @ekamuktia. it will reduce repetition. and might be we can add config/props to exclude page which doesn't need the button.

resir014:
Yep, let's go with this. Please make the necessary changes @kevinmingtarja.

kevinmingtarja:
> good idea @ekamuktia. it will reduce repetition. Might be we can add config/props to exclude page which doesn't need the button.

In order to be able to exclude some pages, should we add the component in `_app.tsx` instead? So that we can access the router props.

resir014 (approved):
🚢

PR #284, opened 2021-07-21
script to generate sitemap index

### LINKED ISSUE #234: SEO - Generate a dynamic sitemap index
## Description

We have multiple Sitemap generated by `next-sitemap`, it's better if we have the Sitemap index that list all available Sitemap on one Sitemap called a sitemap index.

The current workaround is we increase the Sitemap size to reduce the number of Sitemap, and we submit it manually through the Google Search Console. This is not ideal, if the number of Sitemap is increasing.

Sadly, it can not be automatically done on next-sitemap configuration, related with open issue on https://github.com/iamvishnusankar/next-sitemap/issues/69

## Implementation Model

My proposal is we create a custom script after the Sitemap generation. Read all the available Sitemap on the public dir, write new file: `sitemap-index.xml` contains all the Sitemap data.

Use node.js code, utilize the File System API, `fs` from Node.js. You can walk the dir, read all files, and write new file.

Run the script after next-sitemap finished. Maybe modify the [postbuild](https://github.com/kawalcovid19/wargabantuwarga.com/blob/main/package.json#L6) script.

```json
{
  "postbuild": "next-sitemap && node ./your-script.js"
}
```

## Sample XML format

```xml
<?xml version="1.0" encoding="UTF-8"?>
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
   <sitemap>
      <loc>http://www.example.com/sitemap1.xml.gz</loc>
      <lastmod>2004-10-01T18:23:17+00:00</lastmod>
   </sitemap>
   <sitemap>
      <loc>http://www.example.com/sitemap2.xml.gz</loc>
      <lastmod>2005-01-01</lastmod>
   </sitemap>
</sitemapindex>
```

Read more about this topic on https://developers.google.com/search/docs/advanced/sitemaps/large-sitemaps?hl=id

## Priority

Now this ticket is not urgent, since we already done with the manual way.
So maybe we can pick this ticket after got a free space.

ekamuktia:
Closes #234 

## Description

Read sitemap available in public directory, create a `sitemap-index.xml` containing the sitemap list

## Current Tasks
- [x] Create script to generate sitemap index 
- [x] Update `postbuild` to execute the script after sitemap is generated by `next-sitemap`
- [x] Test build / Check if sitemap-index.xml is created after deploy
- [x] Add screenshot
https://deploy-preview-284--wargabantuwarga.netlify.app/sitemap-index.xml
image


etc/create-sitemap-index.ts
```diff
@@ -0,0 +1,40 @@
+import fs from "fs";
+import globby from "globby";
+import path from "path";
+import prettier from "prettier";
+
+const getDate = new Date().toISOString();
+
+const webrootDomain = "https://www.wargabantuwarga.com";
```
mazipan, line 8: We have `~/constants` dir contains `PUBLIC_PATH`, can we use it?

etc/create-sitemap-index.ts
```diff
@@ -0,0 +1,40 @@
+import fs from "fs";
+import globby from "globby";
```
mazipan, line 2: I think `globby` is not native Node.js, have you add the dependency to the package.json?

mazipan (approved):
Rock Solid 💪

PR #286, opened 2021-07-22
Fix search issue

### LINKED ISSUE #231: Search issue in Province Page
## Description

Just like previous bug in FAQ page #203, the same bug still exists on `Province` page.
For example:
1. Search "medika" in https://www.wargabantuwarga.com/provinces/dkijakarta-1
2. Select "Rumah Sakit" for Kategori filter
3. Type "Cahaya Gas" in keyword input
4. Click "cari" button
image

image
The category filters are updated based on the keyword search, but the selected filter on the form is still "Rumah Sakit" which doesn't exist in new category filters so Kategori doesn't have selected item. 
Or in another case like in #203, different filter is selected (probably based on the order on previous filter option?)

The result is also empty because it applies both keyword and the filter (searching for "Cahaya Gas" keyword with "Rumah Sakit" category.

## Expected Behaviour

When `cari` button is clicked, the filter should be cleared/reset(?)
I'm still not sure what behavior will be more natural. 
When user already select a category and search for another keyword:
- Does user looking for a completely new result based on keyword? or
- User wants to search those keyword for items under selected filter?
After the rule is fixed, then the bug should be fixed accordingly.

ekamuktia:
Closes #231

## Description

Based on test case mentioned on #231
Attempt 1: 
Assuming user wants a completely new result based on keyword when `cari` button is clicked, then filter value needs to be reset

Another issue,
On FAQ, I think `kategori_pertanyaan` needs to be added on search fields because for example if you search `isoman` the faq items under `Isoman` category is not included on the result.

## Current Tasks
- [x] Reset filter value on `cari` button clicked
- [x] Add `kategori_pertanyaan` as searchable field for FAQ page
- [ ] Review whether the interaction is better/more natural - Also for FAQ page, the filter value is now cleared when `cari` button is clicked. Check whether to keep the filter value if `checkDocSize` is set as `false`.


pages/faq.tsx
```diff
@@ -36,7 +36,7 @@ export default function Faqs(props: FaqsProps) {
   const [filteredQuestions, handleSubmitKeywords, urlParams, filterItems] =
     useSearch(
       faq,
-      ["pertanyaan", "jawaban"],
+      ["kategori_pertanyaan", "pertanyaan", "jawaban"],
```
mazipan, line 39: My personal take, if you have arguments more than 3, wrap it into an object.
It will be easier for your client to know what is the argument for.

resir014 (approved):
✅ All good from me on code quality side, QA also passed based on all the reproduction steps in the issue. Thanks @ekamuktia!

mazipan (approved):
Looks good ✅

PR #305, opened 2021-07-22
feat: install bundleAnalyzer

### LINKED ISSUE #256: Install BundleAnalyzer
## Description

BundleAnalyzer can help us to know the composition of the JS chunk we have.

With it, we will know if have a heavy 3rd parties, wrong importing strategy, etc

## Tasks

- [ ] Install BundleAnalyzer
- [ ] Add a new script on `package.json` to run this analyzer

mazipan:
Closes #256

## Description

Install BundleAnalyzer

## Current Tasks

- [x] Install new devDeps `@next/bundle-analyzer`
- [x] Add new script `yarn analyze` to run BundleAnalyzer

Screen Shot 2021-07-22 at 09 56 56


package.json
```diff
@@ -5,6 +5,7 @@
     "build": "next build",
     "postbuild": "next-sitemap && ts-node etc/create-sitemap-index.ts",
     "dev": "next",
+    "analyze": "ANALYZE=true yarn build",
```
resir014, line 8:
```suggestion
    "analyze": "cross-env ANALYZE=true yarn build",
```

This will fail to run on Windows. Please use `cross-env` for a cross-platform solution for setting environment variables.

resir014 (changes requested):

package.json
mazipan: Added

resir014 (approved):

PR #306, opened 2021-07-22
feat: try lh ci to replace Netlify Plugin

### LINKED ISSUE #198: Replace Netlify Lighthouse
## Description

The main problem when we put the performance test on the build process is that it will slow down our build time. So the idea is to differentiate the process.

Move the performance check from build time to the PR creation without blocking the build process. It block the PR, not the build process.

## Proposal

Try to check Lighthouse-CI on GitHub Action: https://github.com/GoogleChrome/lighthouse-ci

mazipan:
Closes #198

## Description

Install Lighthouse-CI

It will run on the pull-request with adding `need-perf-check` label.

For main branch, it will automatically triggered in every push

## Current Tasks

- [x] Disable Netlify Plugin for temporary
- [x] Activate LH CLI
- [x] Set a perf budget

Screen Shot 2021-07-22 at 12 07 18

Sample report https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1626930423991-70724.report.html

## Known issues

- Will not creating any comments or Github status check on the PR, we will rely on the Github Action status
- The action will failed when the score didn't pass the gate (you can check on the `lighthouserc.js`)

Screen Shot 2021-07-25 at 17 34 15

## Benefits

- We will have a detailed LH report, just open the action logs, visit the report URL on the log ([sample](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1626930423991-70724.report.html))
- We can have a more complete assertion, e.g Core Web Vitals
- We can pick which PR should be checked by LH


mazipan:
Need to install App https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md#github-app-method-recommended

zainfathoni:
App installed and `secrets.LHCI_GITHUB_APP_TOKEN` set Mas @mazipan. Thanks for working on it!

mazipan:
Screen Shot 2021-07-22 at 14 20 06

Still failed to get the TOKEN, am I wrong? Can we test adding a random secret, e.g: `TEST_TOKEN=THIS_IS_TEST_TOKEN` 

@zainfathoni

mazipan:
Or do we need to try create a `PERSONAL_TOKEN` that have access to create a status check on the commit @zainfathoni?

mazipan:
Or, is this because we use a forked repo ya, so it will not use the secret from this repo?

zainfathoni:
Yes, I think so.

If we put it into the **Environment Variables** instead, can forked repositories access `env. LHCI_GITHUB_APP_TOKEN`?

mazipan:
help to check this @zainfathoni 

https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/disabling-or-limiting-github-actions-for-a-repository#configuring-required-approval-for-workflows-from-public-forks

Setting an approval for first-time contributor only

mazipan:
Asking help from @pveyes 

I create two workflow, the first one is only for trigger, save an artifact to be used later.
But the second workflow can not triggered :-(

What I am missing ya?

pveyes:
IIRC GitHub doesn't allow you to respond to another workflow run by another action because it might trigger infinite loop.

You can only respond on manually-run workflow

mazipan:
Got from this

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

.github/workflows/lighthouse-ci.yml
```diff
@@ -0,0 +1,45 @@
...
+          site_name: 'wargabantuwarga'
+
+      - name: "Lighthouse CI assertion"
+        id: lhci
+        shell: bash
+        run: |
+          yarn global add @lhci/cli@0.8.x
+          echo -e "\n"
+          echo "Start collecting LH report..."
+          lhci collect --url=${{ steps.netlify.outputs.url }} -n=1
```
zainfathoni, line 36: Instead of using the deploy preview URL that is injected with some custom scripts by Netlify, can we get the original URL (the unique hash for each deployment instance)?

Please refer to [this comment ](https://github.com/kawalcovid19/wargabantuwarga.com/pull/19#issuecomment-877888775) for more context.

zainfathoni (changes requested):

.github/workflows/lighthouse-ci.yml
zainfathoni: Hmm. After looking at [the source code here](https://github.com/JakePartusch/wait-for-netlify-action/blob/d85a22dfa96b3f4537d8abdf2dab4ba2a6ac2079/index.js#L32), I don't think it's possible. 😅

Then let's go with this solution, for now, at least it can make the build on Netlify faster. 😁

zainfathoni (approved):
LGTM 💯 

Thanks for working on it, Mas @mazipan! 🙏

PR #309, opened 2021-07-22
Fix same-page anchor reference is hidden behind the header

### LINKED ISSUE #298: Same-page anchor reference is hidden behind the header
## Description

Our current implementation for #285 is not perfect, because when the user clicks the anchor link, the page scrolls to the element with the `id` and putting it at the very top of the page. It makes the title is hidden beneath the header.

As illustrated in this video:

https://user-images.githubusercontent.com/6315466/126579891-e32f2376-55ee-4b4a-b9a0-a35ac4f12a04.mov

## Expected Behaviour

The page should scroll to the section while still showing the title right below the header.

## Additional Information

I solved the same issue in my https://www.ninjavan.co website by adding a dummy element on top of the heading element and style it a few `rem` above the actual heading element.

ekamuktia:
Closes #298 

## Description

Use css to add pseudo element and position it above the actual element with `id` attribute

## Current Tasks
- [x] Update css 
- [x] Check on different browsers/device

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -56,7 +56,8 @@ img {
   display: none;
 }
 
-.home-page-content [id]:before {
+.home-page-content [id]:before,
+.prose [id]:before {
   content: "";
   display: block;
   position: relative;
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -56,6 +56,15 @@ img {
   display: none;
 }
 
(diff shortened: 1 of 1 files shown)


resir014:
@ekamuktia Don't forget to fix the anchor reference on the future homepage as well at `/home-page`.

mazipan (approved):

PR #367, opened 2021-07-23
Add proper title attribute on the province & contact link for SEO improvement

### LINKED ISSUE #334: Add proper title attribute on the link for SEO improvement
## Description
Title attribute can be used to describe html element that's beneficial for SEO. 
Currently, there's no `title` on the internal link, for example on this Province page

image

## Implementation Model

Add title attribute on the link on Province List https://www.wargabantuwarga.com/provinces and contact list for each province
To make it easier, for first iteration we can just use the same value as each page meta title

ekamuktia:
Closes #334

## Description

Create function to generate meta title for province & contact page so it can be reused to set title on the contact & province list.
Also on contact meta title generator, check whether location and province name is the same to prevent duplicate word with same meaning.

## Current Tasks

- [x] Separate meta title generator from getMeta 
- [x] Update meta title generator for contact page to check whether province name and location is the same
- [x] Use meta title generator value for contact & province link title attribute


components/contact-list.tsx
```diff
@@ -4,6 +4,7 @@ import { ContactListSkeleton } from "~/components/ui/skeleton-loading";
 import { anchorTransformer } from "~/lib/htmr-transformers";
 import { Contact } from "~/lib/provinces";
 import { isNotEmpty, stripTags } from "~/lib/string-utils";
+import { getContactMetaTitle } from "~/pages/provinces/[provinceSlug]/[contactSlug]";
```
mazipan, line 8: This is weird, the components require the pages.

mazipan: I think, we can create utilities under `~lib/meta`
Then move to the `getMetaProvince` and the `getMetaContactDetail` maybe 🤔

mazipan (approved):
Looks 🐐

Issue #374, opened 2021-07-24
Migrate to typesense as search backend


famasya:
## Overview
As discussed at #290, we can move to typesense to gain instant-search capability. The search experience should be easier and faster for the user, as it provides typo-tolerant and near real-time search results.

### Implementation details
Typesense server will hosted by trustmedis and its contents are crawled hourly by [wbw-gsheets-crawler](https://github.com/kawalcovid19/wbw-gsheets-crawler). As discussed by @zainfathoni , there are three indices for now:
- `wbw-gsheets-database`
- `wbw-gsheets-faq`
- `wbw-gsheets-telemedicine`

More indices can be added by making MR into crawler repository.

Fortunately, we have [typesense-instantsearch-adapter](https://github.com/typesense/typesense-instantsearch-adapter) as a wrapper of Algolia's [instantsearch react components](https://github.com/algolia/react-instantsearch/). No need to write components from scratch.

Here's information of API KEY
<details><summary>CLICK ME</summary>
<pre>
HOST : public-api.trustmedis.id
PROTOCOL : https
PORT : 443
API KEY : FByczfHEjsTCihgkkYlq2YbAgUKMoyVP
<pre>
</details>

A dirty implementation to instantiate typesense class should be looked like:
```javascript
  const searchAdapter = new TypesenseInstantSearchAdapter({
    server: {
      apiKey: API_KEY, // search-only API key
      nodes: [
        {
          host: "public-api.trustmedis.id",
          port: "443",
          protocol: "https",
        },
      ],
    },
    additionalSearchParameters: {
      queryBy: ["pertanyaan" , "jawaban"], // for wbw-gsheets-faq index
    },
  });
```

## Current Tasks

- [ ] Develop proof-of-concept of searchbox using typesense (TS)
- [ ] Make TS components to be used across pages
- [ ] Implement TS components into "pusat data" page
- [ ] Implement TS components into "FAQ" page
- [ ] Implement TS components into "Telemedicine" page

mazipan:
Just my two cents, the priority should be reordered.
Do not try to create a reusable component on the first place.
It will come into unnecessary complexity.

Just create for spesific page first, then we can think about centralized later.

zainfathoni:
> Just my two cents, the priority should be reordered.
> Do not try to create a reusable component in the first place.
> It will come into unnecessary complexity, an early optimization.
> 
> Just create for any specific page first, then we can think about centralized later.

[...]

PR #384, opened 2021-07-24
simplify search + filter UI

### LINKED ISSUE #244: Simplify search + filter UI
Blocks #134 

## Description

Currently our search + filter UI is a bit messy, and takes up a huge part of the screen. Rework our search UI so that it takes up less space at the top, and allow us to implement something like #134 with a better UX.

## Implementation Model

We can use an approach made by GitHub's mobile web on how they arrange their search + filter UI for issues. We can hide all the filters behind a popup where all the detailed filter options can live.

https://user-images.githubusercontent.com/6315466/126420178-2ed315fc-ee12-476d-a77f-621ee8c6685b.MOV

Dialog state can be controlled by [Headless UI](https://headlessui.dev/react/dialog). ([bundlephobia](https://bundlephobia.com/package/@headlessui/react@1.3.0))

## Tasks

- [ ] Implement filter dialog in Search UI
- [x] #347

resir014:
Closes #244

## Description

Adds a new UI for the search and filter form.

**Mobile**

image

**Desktop**

image

## Current Tasks

- [x] Redesign search + filter UI
- [ ] Debug slow response time for modal


resir014:
So I'm running into some issue here. We already have the filter modal ready, but for some reason if we click the "Filter" button, the modal takes quite a long time to show up, but when I try to dismiss it, it does so instantly.

Here's a demo captured from my stream: https://clips.twitch.tv/TsundereLuckyZebraKappaClaus-A3I-Zmjk0DBi2H4m

Anyone have any idea why is this happening? Is this due to the apropcalypse situation in `SearchForm`?

@zainfathoni

zainfathoni:
> So I'm running into some issue here. We already have the filter modal ready, but for some reason if we click the "Filter" button, the modal takes quite a long time to show up, but when I try to dismiss it, it does so instantly.
> 
> Here's a demo captured from my stream: https://clips.twitch.tv/TsundereLuckyZebraKappaClaus-A3I-Zmjk0DBi2H4m
> 
> Anyone has any idea why is this happening? Is this due to the apropcalypse situation in `SearchForm`?
> 
> @zainfathoni

I tried profiling the performance, and the flame chart looks pretty deep due to `lunr` function (used by `itemsjs` I guess). Therefore, I think one way of optimizing it is by calling whatever `itemsjs` is doing upfront before rendering the modal.

Screenshot 2021-07-25 at 07 42 04

However, the biggest culprit here is the `ContactList` component, so we definitely need to optimize the way it renders, maybe by memoizing some expensive values.

Screenshot 2021-07-25 at 07 44 07

zainfathoni:
After some memoization here and there, I managed to reduce the first-load time of the modal from ~571ms to ~420ms

Before:
Screenshot 2021-07-25 at 08 37 15

After:
Screenshot 2021-07-25 at 08 35 10

and for subsequent load time, it's significantly reduced into ~193ms:

Screenshot 2021-07-25 at 08 35 16

I think we're good to go for now.

zainfathoni (approved):
For the first iteration, it looks good to me. 💯

When we eventually migrate to Typesense at #374, the performance must be improved because we're not relying on `itemsjs` anymore and we can preload the filtering options upfront before the user clicks the `Filter` button.

zainfathoni:
[In production](https://www.wargabantuwarga.com/provinces/dkijakarta-1), it doesn't seem that bad, though. The first load was ~221ms, and the subsequent load was ~86ms.

Screenshot 2021-07-25 at 09 34 21
Screenshot 2021-07-25 at 09 34 34

zainfathoni:
However, in the heavier provinces such as `Jawa Timur`, where it has 400+ entries, it is definitely slower.
First load: ~422ms
Subsequent load: ~107ms

Screenshot 2021-07-25 at 09 36 57
Screenshot 2021-07-25 at 09 37 09

PR #387, opened 2021-07-24
update WBW Opengraph image to new version

### LINKED ISSUE #352: Update the image in meta and opengraph tags
@resir014 we need another version of this banner with the text to be set as the image in the meta tag and opengraph.

_Originally posted by @zainfathoni in https://github.com/kawalcovid19/wargabantuwarga.com/issues/351#issuecomment-885531495_

resir014:
Closes #352

## Description

Updates the WBW Opengraph image to the one similar to the current homepage header.


ekafyi (approved):
LGTM 🚀

PR #390, opened 2021-07-24
Add home image banner field to Netlify CMS

ekafyi:
Closes #341

Accidentally closed PR #364 due to deleted branch, recreate the PR here

## Description

Enable image source URL to be updated from Netlify CMS as text fields. Currently only for the Home Banner image, but can be replicated for other images should we need it.

PR #399, opened 2021-07-25
make banner image path not required

resir014:
## Description

Fixes a regression in Netlify CMS after #390 where the banner image is not meant to be required.

image


resir014:
Duplicate of #398.

PR #401, opened 2021-07-25
Migrate FAQ page into the Typesense backend

### LINKED ISSUE #377: Migrate `FAQ` page into the Typesense backend

ekamuktia:
Closes #377

## Description

Using `react-instantsearch-dom`, creating custom components for `SearchBox`, `Hits`, and `RefinementList`

## Current Tasks

- [ ] Create custom components for `SearchBox`, `Hits`, and `RefinementList`
- [ ] Change the interaction to this form
image
- [ ] Update url query parameters based on search keyword/filter

### DIFF
--- a/lib/typesense.ts
+++ b/lib/typesense.ts
@@ -1,0 +1,27 @@
+import TypesenseInstantSearchAdapter from "typesense-instantsearch-adapter";
+
+export function typesenseSearch({
+  queryBy,
+  defaultSort,
+}: {
+  queryBy: string[];
+  defaultSort?: string;
+}) {
+  const searchAdapter = new TypesenseInstantSearchAdapter({
+    server: {
+      apiKey: "FByczfHEjsTCihgkkYlq2YbAgUKMoyVP", // Be sure to use the search-only-api-key
+      nodes: [
+        {
+          host: "public-api.trustmedis.id",
+          port: "443",
+          protocol: "https",
+        },
+      ],
+    },
+    additionalSearchParameters: {
+      queryBy: queryBy.join(),
+      sortBy: defaultSort ? defaultSort : "",
+    },
+  });
+  return searchAdapter.searchClient;
+}
--- a/@types/typesense-instantsearch-adapter/index.d.ts
+++ b/@types/typesense-instantsearch-adapter/index.d.ts
@@ -1,0 +1,31 @@
+declare module "typesense-instantsearch-adapter" {
+  type SearchClient = object;
+
+  export interface TypesenseNode {
+    host: string;
+    port: string;
+    protocol: string;
+  }
+
+  export interface TypesenseSearchParameters {
+    queryBy: string;
+    sortBy?: string;
+    highlightFullFields?: string;
+  }
+
+  export interface TypesenseServer {
+    apiKey: string;
+    nodes: TypesenseNode[];
+  }
+
+  export interface TypesenseInstantsearchAdapterOptions {
+    server?: TypesenseServer;
+    additionalSearchParameters: TypesenseSearchParameters;
+  }
+
+  export default class TypesenseInstantsearchAdapter {
+    readonly searchClient: SearchClient;
+
+    constructor(options: TypesenseInstantsearchAdapterOptions);
+  }
+}
--- a/components/faq-list.tsx
+++ b/components/faq-list.tsx
@@ -1,0 +1,80 @@
+import { FaqData } from "~/lib/faq-databases";
+
+import htmr from "htmr";
+
+type FAQListProps = {
+  data: FaqData[];
+};
+
+function groupBy<T, U>(data: T[], key: U) {
+  return data.reduce((acc: any, currentValue: any) => {
+    const groupKey = currentValue[key];
+    if (!acc[groupKey]) {
+      acc[groupKey] = [];
+    }
+    acc[groupKey].push(currentValue);
+    return acc;
+  }, {});
+}
+
+export function FAQList(props: FAQListProps) {
+  const listFaqs = groupBy(props.data, "kategori_pertanyaan");
+  return (
+    <div className="space-y-4">
+      {Object.keys(listFaqs as Record<string, unknown>).map(
+        (category: string) => (
+          <div
+            key={category}
+            className="p-4 bg-white shadow overflow-hidden rounded-md"
+          >
+            <div className="relative">
+              <div
+                aria-hidden="true"
+                className="absolute inset-0 flex items-center"
+              >
+                <div className="w-full border-t border-gray-300" />
+              </div>
+              <div className="relative flex flex-row items-center justify-start">
+                <span className="pr-3 bg-white text-lg font-medium text-gray-900">
+                  {category}
+                </span>
+              </div>
+            </div>
+            <dl className="divide-y divide-gray-200">
+              {listFaqs[category].map((question: FaqData) => (
+                <div
+                  key={question.pertanyaan}
+                  className="pt-6 pb-8 md:grid md:grid-cols-12 md:gap-8"
+                >
+                  <dt className="text-base font-semibold text-gray-900 md:col-span-5">
+                    {question.pertanyaan}
+                  </dt>
+                  <dd className="mt-2 md:mt-0 md:col-span-7">
+                    <p className="text-base text-gray-500">
+                      {htmr(question.jawaban)}
+                    </p>
+                    <small>
+                      Sumber:{" "}
+                      {question.link ? (
+                        <a
+                          className="underline text-blue-800"
+                          href={question.link}
+                          rel="noreferrer"
+                          target="_blank"
+                        >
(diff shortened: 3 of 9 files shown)


famasya:
Neat! 

Should we highlight relevant words in the search results? I think it will provide clarity for the user, as it gives context why presented results are relevant.

image

resir014:
@ekamuktia After some discussions with @zainfathoni earlier, we decided to go forward with using Typesense, since the current search infrastructure gave us more problems than expected (e.g. #405). Therefore we need to migrate to Typesense sooner than expected.

Let's see if we can get this up and running asap as a proof-of-concept, so we can apply it to other places (e.g. province search).

ekamuktia:
@resir014 @zainfathoni 
The UI itself is customizeable, so it doesn't matter whether we use original typesense widget (https://www.algolia.com/doc/guides/building-search-ui/widgets/showcase/react/) or customize it to current UI
On my local development, the problem is solved but somehow on the deploy preview the handleFilterChange doesn't work. 
Changed it to non modal filter because I thought it's probably the problem, but same thing still happens.
Any idea why?

Local:

https://user-images.githubusercontent.com/9606523/126901072-d4c92172-9b4e-47be-ab77-45c224048a29.mp4

Deploy Preview:

https://user-images.githubusercontent.com/9606523/126901096-84c356bb-b3db-41ea-aded-b8919ffc853a.mp4

components/search/custom-instant-search.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+type FilterSetting = {
+  field: string;
+  title: string;
+};
+
+interface CustomInstantSearchProps extends InstantSearchProps {
+  itemName: string;
+  filterSettings?: FilterSetting[];
+  useFilterModal?: boolean;
```
resir014, line 30:
```suggestion
  withFilterModal?: boolean;
```

`useFilterModal` -> `withFilterModal` - since functions starting with `use` is a common convention for React hooks, we don't want to confuse this prop with a hook. 😅

resir014 (changes requested):

lib/typesense.ts
```diff
@@ -0,0 +1,27 @@
...
+export function typesenseSearch({
+  queryBy,
+  defaultSort,
+}: {
+  queryBy: string[];
+  defaultSort?: string;
+}) {
+  const searchAdapter = new TypesenseInstantSearchAdapter({
+    server: {
+      apiKey: "FByczfHEjsTCihgkkYlq2YbAgUKMoyVP", // Be sure to use the search-only-api-key
```
resir014, line 12: I think we should start storing these secrets on an environment variable. @zainfathoni what do you think?

zainfathoni: Agreed. Just let me or @resir014 know what environment variable to store, Mbak @ekamuktia, we can do it for you. 😁

resir014:
@zainfathoni @ekamuktia Created an environment variable in the Netlify dashboard called `NEXT_PUBLIC_TYPESENSE_API_KEY`. Let's put all secret values (e.g. `apiKey`) there.

zainfathoni:
What about the GitHub Actions build, @resir014? I think we should also put some env vars on the GitHub Secrets, shouldn't we?

ekamuktia:
> @zainfathoni @ekamuktia Created an environment variable in the Netlify dashboard called `NEXT_PUBLIC_TYPESENSE_API_KEY`. Let's put all secret values (e.g. `apiKey`) there.

For this, how can I use it on the code? Like this process.env.NEXT_PUBLIC_TYPESENSE_API_KEY?

@famasya @zainfathoni @resir014 
Finally, I found the problem why the filter didn't work on deploy preview and fixed it.
Some things to note:
- For FAQ, I couldn't find way to sort it based on the document order (tried `objectID:asc` and `id:asc`, but both failed)
- For test case, since the data is from index, then it cannot use mock data. Test case needs to be updated.
- Since I'm using new components, some functionality still not applied here (loading, empty state, preserve url parameter, etc)

famasya:
> > @zainfathoni @ekamuktia Created an environment variable in the Netlify dashboard called `NEXT_PUBLIC_TYPESENSE_API_KEY`. Let's put all secret values (e.g. `apiKey`) there.
> 
> For this, how can I use it on the code? Like this process.env.NEXT_PUBLIC_TYPESENSE_API_KEY?
> 
> @famasya @zainfathoni @resir014
> Finally, I found the problem why the filter didn't work on deploy preview and fixed it.
> Some things to note:
> 
> * For FAQ, I couldn't find way to sort it based on the document order (tried `objectID:asc` and `id:asc`, but both failed)
> * For test case, since the data is from index, then it cannot use mock data. Test case needs to be updated.
> * Since I'm using new components, some functionality still not applied here (loading, empty state, preserve url parameter, etc)

For the sorting, it's intended since the schema I wrote doesn't store any numeric value to be sorted. Results are sorted by relevance by default.

Do you need any sortable field for this? If so, let's put that

zainfathoni:
> > @zainfathoni @ekamuktia Created an environment variable in the Netlify dashboard called `NEXT_PUBLIC_TYPESENSE_API_KEY`. Let's put all secret values (e.g. `apiKey`) there.
> 
> For this, how can I use it on the code? Like this process.env.NEXT_PUBLIC_TYPESENSE_API_KEY?

Yes, something like that.
And to make it work in the GitHub actions, you need to pass `secrets.NEXT_PUBLIC_TYPESENSE_API_KEY` to the environment variables.

ekamuktia:
Tried to use process.env.NEXT_PUBLIC_TYPESENSE_API_KEY but the testing failed so I can't commit/push.
Tried adding `setupFiles: ["dotenv/config"]` in `jest.config.js` as mentioned [here](https://tekloon.dev/using-dotenv-with-jest) but it didn't work.

ekamuktia:
> For the sorting, it's intended since the schema I wrote doesn't store any numeric value to be sorted. Results are sorted by relevance by default.
> 
> Do you need any sortable field for this @ekamuktia ? If so, I'll put that on

@famasya 
For FAQ, I may need it based on the row order on the sheet (just like current order)
And for Contact data, I need to order by `verifikasi`

mazipan:
Putting reference: https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react/

In case we want to support a static filtering via path param

mazipan:
How can I checkout from this branch ya 😂?

zainfathoni:
> How can I checkout from this branch ya 😂?

You can use [GitHub VSCode Extension](https://code.visualstudio.com/docs/editor/github#_reviewing), Mas @mazipan.
There's an option to checkout to the Pull Request, and when you make commits on top of it, you can push it into the remote branch, even if the branch is on a forked repository. 😎 
 
Screenshot 2021-08-13 at 06 48 39

__tests__/pages/faq.test.tsx
```diff
@@ -1,132 +1,120 @@
 import React from "react";
 
-import { perBuild } from "@jackfranklin/test-data-bot";
+// import { faqBuilder } from "~/lib/__mocks__/builders/faq";
+// import faqs from "~/lib/data/faqs";
```
bikkimahato, line 4: Commented code should not be pushed.

__tests__/pages/faq.test.tsx
```diff
@@ -1,132 +1,120 @@
...
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import FaqPage, { getStaticProps } from "../../pages/faq";
-import { faqBuilder } from "~/lib/data/__mocks__/builders/faqs";
-import faqs from "~/lib/data/faqs";
+import FaqPage from "../../pages/faq";
+
+// import { perBuild } from "@jackfranklin/test-data-bot";
```
bikkimahato, line 9: Commented code should not be pushed.

components/search/custom-instant-search.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const urlToSearchState = () => {
+    const searchParams: SearchState = {};
+    if (typeof window !== "undefined") {
+      const queryParams: {} = getQueryParams(window.location.search);
+      if (Object.keys(queryParams).length) {
+        let keywordsParam: string = "";
+        const filtersParam: { [key: string]: string[] } = {};
+        const filterFields = filterSettings?.map((cur) => cur.field) ?? [];
+        Object.entries(queryParams).forEach(([key, value]) => {
+          if (key == "q") {
```
bikkimahato, line 52: Please use strict equality operator ===

components/search/custom-instant-search.tsx
```diff
@@ -0,0 +1,206 @@
...
+            keywordsParam = value as string;
+            if (keywordsParam) {
+              searchParams.query = keywordsParam;
+            }
+          } else if (key == "sort") {
```
bikkimahato, line 57: Please use strict equality operator ===

PR #361, opened 2021-07-23
feat: Add static slug for province detail page 

Namchee:
Closes #288

## Description

Generate static slugs for all provinces and apply all related changes to the client side code.

## Current Tasks

- [x] Generate slug from the etc/fetcher, slugify the title, e.g: DKI Jakarta will be dki-jakarta
- [x] Use the slug on the client side, without any additional index
- [x] Test the functionality

PR #416, opened 2021-07-25
fix: Revert back to mirror-box

### LINKED ISSUE #412: Revert `fetch-wbw` back to `mirror-box` again once the `mirror-box` is updated
> @zainfathoni the netlify build will always fail since I added a new static property `slug` when generating data with `etc/fetchers`. The new format isn't generated when netlify executes `yarn build` which will cause `provinceSlug` to be `undefined` thus causing `next build` to throw errors (as it expects the slug to be at least `null` or empty string).
> 
> Can we make sure that the app fetches fresh data before any build?

You can temporarily revert [this line](https://github.com/kawalcovid19/wargabantuwarga.com/commit/81be7ad788d76a02107e896fc3c14f95b124b89b#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R14) into using the `fetch-wbw` script again, so that we can merge this PR to the `main` branch.
Once it has been merged to the `main` branch, the `mirror-box` script gets the latest version of your `fetch-wbw.js` implementation after a few minutes. After that, you can raise another PR to revert your changes to use the `mirror-box` again. 😅 

I know it's a hassle, but I don't think we need to optimize this workflow, because #374 is coming soon anyway, and it will render both `fetch-wbw` and `mirror-box` useless. 😅

_Originally posted by @zainfathoni in https://github.com/kawalcovid19/wargabantuwarga.com/issues/361#issuecomment-886164028_

Namchee:
Closes #412

## Description

Revert back to `mirror-box` from `fetch-wbw` due to #361.

## Current Tasks

- [x] Replace `fetch-wbw` with `mirror-box`

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "lint": "eslint \"**/*.{js,jsx,ts,tsx}\"",
     "lint:fix": "eslint --fix \"**/*.{js,jsx,ts,tsx}\"",
     "mirror-box": "ts-node etc/mirror-box.ts",
-    "netlify-export": "yarn fetch-wbw && yarn build && next export",
+    "netlify-export": "yarn mirror-box && yarn build && next export",
     "prepare": "husky install",
     "start": "next start",
     "cypress:open": "cypress open",


zainfathoni:
@adityapurwa do you need to manually clone the repository and restart the `pm2` process in your server to make this build successful?

mazipan (approved):

adityapurwa:
> @adityapurwa do you need to manually clone the repository and restart the `pm2` process in your server to make this build successful?

I just need to pull the new changes, no restart is necessary as pm2 will always use the latest code. I updated the server to use this new version. Thank you for letting me know of these changes! 👍

PR #320, opened 2021-07-22
Swap the `index.tsx` with the `home-page.tsx` file

zainfathoni:
Closes #39

## Current Tasks

- [x] Swap the pages
- [x] Update Netlify CMS configuration to link to the new `index.tsx` page
- [x] Manual regression test

PR #423, opened 2021-07-25
revert: bring back gtm

### LINKED ISSUE #419: Google Tag Manager is not triggered in the new homepage
## Description

Previous Google Tag Manager is no longer triggered in Warga Bantu Warga website after the homepage revamps, thus creating an issue where data is not sent to Google Analytics. 

## Expected Behaviour

Google Tag Manager should be triggered on every page, see the previous PR here: https://github.com/kawalcovid19/wargabantuwarga.com/pull/36

## Additional Information

- [x] Add the GTM trigger back to WBW
- [x] Add proper noscript implementation for Google Tag Manager in WBW
- [ ] Add unit test to this critical functionality to ensure that future refactor will be less likely to produce the same bug, or later engineers will notice if this bug is recurring. Perhaps using React Test Library or Enzyme #425

See the discussion thread here: Screen Shot 2021-07-25 at 22 31 49

mazipan:
Closes #419

## Description

Based on this PR #320, I revert the changes about GTM, move it to the top level

## Current Tasks

- [x] Take the code from #320
- [x] Bring back to the top level, not the page level
- [x] Remove double GTM on the `home-page.tsx`


baraeb92:
Is this something that we can add unit test to? 

Just to make sure similar issue less likely happening in the future? @mazipan

mazipan:
I think we can cover this case by our automated testing.
Will create on seperate PR.

Since this is a high priority bug, we need to deliver the fix as fast as we can.

mazipan:
Create #425 for the testing part

resir014 (approved):

pages/_app.tsx
```diff
@@ -82,6 +83,25 @@ export default function App({ Component, pageProps, router }: AppProps) {
         <link href="/manifest.json" rel="manifest" />
         <meta content="#1667C2" name="theme-color" />
       </Head>
+
```
baraeb92, line 86: This line 86 could be removed I think

pages/_app.tsx
```diff
@@ -82,6 +83,25 @@ export default function App({ Component, pageProps, router }: AppProps) {
...
+        }}
+      />
+      <noscript>
+        <iframe
+          height="0"
+          src="https://www.googletagmanager.com/ns.html?id=GTM-5X4ZPBX"
+          style={{ display: "none", visibility: "hidden" }}
+          title="gtm"
+          width="0"
+        />
```
baraeb92, line 103: need to add </iframe> instead for the noscript to work properly. 

See previous implementation here: https://github.com/kawalcovid19/wargabantuwarga.com/pull/36/files?file-filters%5B%5D=.tsx

baraeb92 (commented):
Just some concern about this mas @mazipan @zainfathoni 

Need to make sure that the noscript is also present on the body, currently, when loading the web, it seems that there's no noscript present when I check the source code.

pages/_app.tsx
zainfathoni: in React, they should be identical, Mas @baraeb92.

But somehow it's not showing in the page now, could you please investigate it further, Mas @mazipan?

mazipan: Bisa insert expectation kode komplit nya dari GTM gak ya, Mas @baraeb92.?

mazipan: Since all this code are just `copy-paste` from previous implementation without any modifications

mazipan: Since all this code are just copy-paste from previous implementation without any modifications

mazipan: Screen Shot 2021-07-26 at 10 50 41

It already exist, it just under `main` tag, should we move it under `body`?

PR #393, opened 2021-07-25
refactor Page components, add Database page CTA

resir014:
Closes #391
Closes #406 
Partial work for #333

## Description

Reworking how the Page components work to better implement the Website 2.0 designs, as well as initial steps for the homepage CTA boxes.

image

## Tasks

- [x] Database CTA
- [x] Database menus
- [x] Sembako CTA

PR #433, opened 2021-07-26
Fix: selected filter url param not exist

ekamuktia:
Partial work for #393 

## Description

Fix search filter, use `filter` param instead of `filters` because `filters` only works if the data exists


ekamuktia:
@zainfathoni I can only continue after 4 or 5pm today, so if #393 is urgent you can copy the added part or add the test code here 🙏

resir014 (approved):
Manual QA passed, just need to get the test coverage up to par. @zainfathoni @ekamuktia

zainfathoni:
Finally, it passes the Codecov checks! 🎉

zainfathoni (approved):
LGTM 💯 

Thanks, Mbak @ekamuktia, for fixing this bug.

PR #445, opened 2021-07-26
style: reorder bottom navigation & change wording

### LINKED ISSUE #422: [UX] Bottom Page Navigation Improvements
## Description

there are several things that interfere with the bottom page navigation, including the location of the navigation items that are different from the previous version, and the use of mixed language.

126778941-3cee2218-fb7a-4cff-94a0-412b84f25d8e

`Pusat Data` used more often, but is in third place.

Capture

## Tasks

- [ ] Reorder navigation items according to the updated Figma file
- [ ] Rename menu items according to the updated Figma file.

ihsanrabb:
Closes #422 

## Current Tasks

- [x]  Reorder navigation items according to the updated [Figma file](https://www.figma.com/file/XNNtIoFEdFqaXOee83n0oN/WBW?node-id=485%3A2485)
- [x] Rename menu items according to the updated [Figma file.](https://www.figma.com/file/XNNtIoFEdFqaXOee83n0oN/WBW?node-id=485%3A2485)


resir014 (changes requested):
The ordering is incorrect. Please take a look at the [updated Figma file](https://www.figma.com/file/XNNtIoFEdFqaXOee83n0oN/WBW?node-id=485%3A2485) for reference:

image

ihsanrabb:
Base on @fncolon  description issue #422, Order Navigation Items Like Previous Bottom Page Navigation but it's different in  [updated Figma file](https://www.figma.com/file/XNNtIoFEdFqaXOee83n0oN/WBW?node-id=485%3A2485). Which order should we implement mas @resir014 ?

resir014:
@ihsanrabb Let's follow the Figma. cc: @baraeb92

resir014:
I've also updated the task list in the issue to avoid any confusion.

components/layout/navigation.tsx
```diff
@@ -14,8 +14,8 @@ export function Navigation() {
 
   const menuClasses = (isActive: boolean) => {
     return [
-      "inline-flex flex-col items-center justify-center h-12 px-4 rounded-md",
-      isActive ? "text-blue-600 font-semibold" : "text-gray-600",
+      "inline-flex flex-col items-center justify-center h-12 px-2 rounded-md",
+      isActive ? "text-blue-600 font-medium" : "text-gray-600",
```
resir014, line 18:
```suggestion
      isActive ? "text-blue-600 font-semibold" : "text-gray-600",
```

The design on the Figma file uses a `font-size` of 600, which is `font-semibold`. Please keep it as it is.

image

resir014 (changes requested):

components/layout/navigation.tsx
```diff
@@ -27,13 +27,13 @@ export function Navigation() {
           className: "w-8 h-8",
           "aria-hidden": true,
         })}
-        <span className="text-xs">{item.name}</span>
+        <span className="text-xs text-center">{item.name}</span>
```
resir014, line 30: @ihsanrabb Is this necessary? Didn't `items-center` above already sort the centering?

ihsanrabb: I think it's better to add a `text-center`. `items-center` only center an icon on smaller screen @resir014 
bottom-nav

resir014: Okay. Then I think we should move this up higher up the element tree, alongside `items-center justify-center` @ihsanrabb

resir014 (approved):

PR #452, opened 2021-07-26
perf: improve GTM initialisation

### LINKED ISSUE #439: Replace custom script component with `next/script`
## Description

Our current `<Script />` component implementation has an issue where it would load the same script multiple times.

image

Replace this component with the built-in `next/script` component to fix this issue.

## Implementation Model

[Documentation for `next/script`](https://nextjs.org/docs/basic-features/script)

**Important reminder:** **do not** put `next/script` inside the `_document.tsx` file, as there's some client-side functionality required.

image

## Tasks

- [ ] Replace GTM script to use `next/script`
- [ ] Replace other scripts to use `next/script`

resir014:
Closes #439 

## Description

Replace custom `next/script` implementation, and improve the initialisation of GTM script.


pages/_app.tsx
```diff
@@ -6,11 +6,12 @@ import "typeface-inter";
 import { useEffect } from "react";
 
 import { LayoutRoot } from "~/components/layout/layout-root";
-import { Script } from "~/components/script";
```
zainfathoni, line 9: Please remove the custom `Script` component as well, thanks!

lib/gtm.ts
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-condition */
+export const initializeGTM = () => {
+  window.dataLayer = window.dataLayer || [];
+  window.dataLayer.push({ "gtm.start": new Date().getTime(), event: "gtm.js" });
```
zainfathoni, line 4: Is there any reference on how this could delay the initializations and improve the performance?

zainfathoni (changes requested):
Minor change requests and clarifications.

lib/gtm.ts
resir014: @zainfathoni Since this is called on the `useEffect()` hook, it will follow the hooks lifecycle. No idea on how this will be any faster, but at least this fixes the multiple `<script />` tags issue, since for some reason only inline scripts are affected.

However, this will probably be fixed in a future update to Next.js. (https://github.com/vercel/next.js/pull/27218/files)

package.json
```diff
@@ -18,8 +18,8 @@
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
     "test": "is-ci \"test:coverage\" \"test:watch\"",
-    "test:coverage": "jest --coverage",
-    "test:watch": "jest --watch",
+    "test:coverage": "jest --coverage --env=jsdom",
+    "test:watch": "jest --watch --env=jsdom",
```
zainfathoni, line 22: If you need the `jsdom` environment in the test, please update the regex in the `test/jest.client.js` file to match the file and update the regex in the `test/jest.server.js` file to exclude the file.

zainfathoni (changes requested):

pages/_app.tsx
resir014: already did!

package.json
resir014: Done.

zainfathoni (approved):
LGTM 💯 

Thanks, @resir014!

PR #473, opened 2021-07-27
fix: refactor mobile search form

### LINKED ISSUE #444: Refactor Mobile Search Form
## Description
Just like suggested in #421, to reduce search form space, change mobile search form styling just like desktop version 

Current:

Expected:
Screen Shot 2021-07-25 at 22 02 00

## Tasks

- [x] Move cari button inline with keyword input
- [x] Show 2 filter items per row (2 column grid)

ekamuktia:
Closes #444

## Description
To reduce search form space, change mobile search form styling just like desktop version

Problem found: on original styling, the primary button default size is `md` which has class `text-sm`. This styling causing the keyword input and `cari` button have different height in mobile. 

image

## Current Tasks
- [x] Move `cari` button inline with keyword input
- [x] Change default size of Primary Button as `undefined` to enable `text-base sm:text-sm` styling so it won't affect other buttons and make keyword input & `cari` button have same height in mobile
- [x] Show 2 filter items per row (2 column grid)
- [x] Update test case

### DIFF
--- a/components/ui/button/primary-button.tsx
+++ b/components/ui/button/primary-button.tsx
@@ -26,7 +26,7 @@ export const PrimaryButton = React.forwardRef<HTMLButtonElement, ButtonProps>(
       style,
       type,
       block,
-      size = "md",
+      size,
       color = "blue",
       rounded,
       icon,
--- a/components/ui/button/primary-button.tsx
+++ b/components/ui/button/primary-button.tsx
@@ -26,7 +26,7 @@ export const PrimaryButton = React.forwardRef<HTMLButtonElement, ButtonProps>(
       style,
       type,
       block,
-      size,
+      size = "md",
       color = "blue",
       rounded,
       icon,
--- a/components/ui/button/utils/helpers.ts
+++ b/components/ui/button/utils/helpers.ts
@@ -32,7 +32,7 @@ export function buttonSizes(size?: ButtonSizes) {
       return "px-6 py-3 text-base";
     }
     default: {
-      return "px-4 py-2 text-sm";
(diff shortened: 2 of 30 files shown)


zainfathoni (approved):
LGTM 💯 

Thanks, Mbak @ekamuktia! 🙏

components/ui/button/primary-button.tsx
```diff
@@ -26,7 +26,7 @@ export const PrimaryButton = React.forwardRef<HTMLButtonElement, ButtonProps>(
       style,
       type,
       block,
-      size = "md",
+      size,
```
resir014, line 29: Why is this change necessary?

components/ui/button/utils/helpers.ts
```diff
@@ -32,7 +32,7 @@ export function buttonSizes(size?: ButtonSizes) {
       return "px-6 py-3 text-base";
     }
     default: {
-      return "px-4 py-2 text-sm";
+      return "px-4 py-2 text-base sm:text-sm";
```
resir014, line 35: This shouldn't be changed as this will cause the button to have different height for the button.

image

resir014 (changes requested):

resir014:
Please check again @ekamuktia. In general, the button sizes **shouldn't** be changed.

resir014:
@ekamuktia Decided to fix the form styles myself so all you need to do is revert the button styling.

#474

components/ui/button/utils/helpers.ts
ekamuktia: Reverted

resir014 (approved):

PR #363, opened 2021-07-23
Design System Alert

hanihusam:
Closes #214 

## Description

Create reusable Alert component

## Current Tasks
- [x] Create the base component
- [x] Allow to add icon
- [x] Make it dissmisable
- [x] Testing for the component

PR #487, opened 2021-07-28
feat(ui): added outline button + `light-blue` colours

### LINKED ISSUE #479: Add new Outline Button component
## Description

Add a new Outline button component to our Button design system components. Outline Buttons are buttons with transparent background but with coloured texts + borders.

image

image

## Implementation Model

- Create new files `outline-anchor-button.tsx` and `outline-button.tsx`.
  - Similar props, dimensions, and structure as the other buttons.
  - Only border colours. Background should be transparent, but `-100` when hovered.
- Add tests for these buttons.

resir014:
Closes #479

## Description

Added `OutlineButton` + `OutlineAnchorButton`, as well as an additional `light-blue` colours for future components.

Here's how it'll look like:

image


zainfathoni:
What do you think is causing the CI errors, @resir014? If I remember correctly, this is not the first instance of the same error.

resir014:
@zainfathoni This might just be the GitHub Actions runner acting up. I usually just re-run all jobs.

zainfathoni (approved):
LGTM 💯 

Thanks, @resir014! 🙏

resir014:
Everything I do, it's always not enough. 💩 

image

zainfathoni:
Maybe you should do the same thing I did [here](https://github.com/kawalcovid19/wargabantuwarga.com/pull/363/commits/ed16d94220b81d27b1fd20bd4691c88c7f9fbdcf).

PR #493, opened 2021-07-28
feat(ui): add `iconPosition` prop to button

### LINKED ISSUE #480: Add `iconPosition` prop in Button design system components
## Description

Add an `iconPosition` prop in all Button components with an option of `"left"` and `"right"`.

image

## Implementation Model

See [Tailwind UI](https://tailwindui.com/components/application-ui/elements/buttons#component-33ea45530bf476032a08474a765a0d2f) for reference.

- Available options: `"left"` and `"right"`
- Default: `"left"`

resir014:
Closes #480

## Description

Add an `iconPosition` prop to button which defaults to `"left"`

### DIFF
--- a/components/ui/button/utils/helpers.ts
+++ b/components/ui/button/utils/helpers.ts
@@ -88,7 +88,7 @@ export function buttonIconClasses(
     case "xl": {
       return [
         iconPosition === "right" ? "ml-3 -mr-1" : "-ml-1 mr-3",
-        "-ml-1 mr-3 h-5 w-5",
+        "h-5 w-5",
       ];
     }
     default: {
--- a/components/ui/button/utils/helpers.ts
+++ b/components/ui/button/utils/helpers.ts
@@ -1,17 +1,24 @@
 import * as React from "react";
 
-import { ButtonColors, ButtonProps, ButtonSizes } from "./types";
+import {
+  ButtonColors,
+  ButtonIconPositions,
+  ButtonProps,
+  ButtonSizes,
+} from "./types";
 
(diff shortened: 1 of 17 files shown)


components/ui/button/utils/helpers.ts
```diff
@@ -186,14 +214,20 @@ export function outlineButtonColors(color?: ButtonColors) {
...
+  icon,
+  size,
+  additionalClasses,
+  iconPosition = "left",
+}: {
+  icon?: ButtonProps["icon"];
+  size?: ButtonSizes;
+  additionalClasses?: string;
+  iconPosition?: ButtonIconPositions;
+}) {
```
resir014, line 227: Changed these parameters to a single object since we're already over 3 parameters now.

components/ui/button/utils/helpers.ts
```diff
@@ -1,17 +1,24 @@
...
-  }
-
-  return "inline-flex";
+export function buttonBlockStyles(
+  block?: boolean,
+  iconPosition: ButtonIconPositions = "left",
+) {
+  return [
+    block ? "flex" : "inline-flex",
+    iconPosition === "right" ? "flex-row-reverse" : "flex-row",
```
resir014, line 20: Don't know if doing this is better than reversing the rendering in the component itself, but this reduces the chance of code copy/paste. Though I guess converting `renderButtonIcon` to a general-purpose `renderButtonContent` might also work, but...

mazipan (approved):
Looks solid 🚀

PR #536, opened 2021-08-01
fix: use correct SEO title + improve a11y on Isolasi Mandiri page

resir014:
### One.

Fixed the SEO title in Isolasi Mandiri page so the site name doesn't get duplicated. (by default the title template set by next-seo is `%s | site_name`)

**Before**

image

**After**

image

### Two.

Improve accessibility on Isolasi Mandiri page + its stacked links by using proper semantic HTML elements instead of `div`.


mazipan (approved):
Looks good to me

PR #538, opened 2021-08-01
perf(seo): change Twitter Card handle

resir014:
## Description

WBW now has an official Twitter handle, so we change the Twitter Card's `handle` setting to that.


mazipan (approved):
Looks good

PR #541, opened 2021-08-01
Fix internal links in Isolasi Mandiri page opens in a new tab

### LINKED ISSUE #535: Internal links in Isolasi Mandiri page opens in a new tab
## Description

On the stacked links in Isolasi Mandiri page, links that are meant to be internal opens in a new tab instead of in the current one.

image

## Expected Behaviour

Internal links (`e.g. FAQ`) should be opened in the same tab.

## Additional Information

The current implementation on `StackedLink` makes all links always open on a new tab, regardless of whether it is an external link or not. We should add a way to differentiate this, either through an extra Boolean item on the Netlify CMS page, or through a logic within the component.

ekamuktia:
Closes #535 

## Description

Check whether link is internal link by checking if the domain name contains "wargabantuwarga.com" (subdomain is also recognized as internal link) or whether it starts with `/` if the link is without domain name

## Current Tasks

- [x] Add internal link checking function 
- [x] Update StackedLink to open internal link in current tab
- [x] Add test cases


lib/__tests__/string-utils.test.ts
```diff
@@ -92,3 +93,23 @@ describe("allIsEmptyString", () => {
...
+
+describe("isInternalLink", () => {
+  it.each`
+    input                                                                     | expected
+    ${"/faq"}                                                                 | ${true}
+    ${"https://www.wargabantuwarga.com"}                                      | ${true}
+    ${"https://www.wargabantuwarga.com/"}                                     | ${true}
+    ${"https://www.wargabantuwarga.com/provinces"}                            | ${true}
+    ${"https://www.wargabantuwarga.com/provinces"}                            | ${true}
+    ${"https://hotline.wargabantuwarga.com"}                                  | ${true}
```
mazipan, line 105: I think `https://hotline.wargabantuwarga.com` is not internal, it different subdomain.
It's safe to throw to new tab

ekamuktia: Updated. Let me know if it should be able to check link with no protocol (//www.wargabantuwarga.com), non http/https protocol, anchor tag  or other additional conditions.

lib/string-utils.ts
```diff
@@ -113,3 +113,15 @@ export function getQueryParams(query: string): {} {
...
+
+export function isInternalLink(link: string): boolean {
+  const domain = link
+    .replace("https://", "")
+    .replace("http://", "")
+    .split("/")[0];
+  return (
+    link.startsWith("/") ||
+    domain == "www.wargabantuwarga.com" ||
+    domain == "wargabantuwarga.com"
```
resir014, line 125: Always use `===`.

components/layout/stacked-link.tsx
```diff
@@ -29,7 +30,7 @@ export default function StackedLink(list: StackedListProps) {
               <a
                 className="helper-link-cover"
                 data-testid={`next-link-${link.title}`}
-                target="_blank"
+                target={isInternalLink(link.url) ? "_self" : "_blank"}
```
resir014, line 33: This does nothing when the `Link` component still wraps the entire `<a>` tag, which will override whatever's made here.

We need to modify this element to be wrapped with `Link` or not based on whether or not the URL is internal or external. See [here](https://github.com/kawalcovid19/wargabantuwarga.com/blob/main/components/layout/navigation-menu.tsx#L68-L84) for example.

resir014 (changes requested):
This need more work, please take a look at the comments.

mazipan (approved):
Looks good to me

resir014 (approved):

lib/string-utils.ts
zainfathoni: Shall we enforce this rule using ESLint, @resir014?

PR #562, opened 2021-08-04
fix: text overflow in link

### LINKED ISSUE #224: Text Overflow in Link
## Description

image

## Expected Behaviour

the link should not be overflowing the container

## Additional Information

it might be better to use another approach like implementing layout as column rather than row to accomodate long text

abui-am:
Closes #224

## Description

Long link caused minor visual bug, this happens because we can't break down a text without space causing it to break out of the container. I tried to create a copy box as suggested in the discussion, hope it help.

## Copy Box That I Created
image

## Current Tasks

- [x] Create copy box
- [x] Add copy button


styles/globals.css
```diff
@@ -66,6 +66,18 @@ img {
...
+.copybox {
+  scrollbar-width: none; /* Firefox */
+  -ms-overflow-style: none; /* IE 10+ */
+}
+
+.copybox::-webkit-scrollbar {
+  width: 0;
+  height: 0;
+  background: transparent; /* Chrome/Safari/Webkit */
+}
```
resir014, line 79: Please wrap this class inside `@layer components` so it will be properly purged on pages that don't use this class.

Also nitpick: `.copybox` -> `.helper-scroll-overflow` - use a more general-purpose name for utility classes.

components/contact-details.tsx
```diff
@@ -29,6 +29,7 @@ type DescriptionItemProps = {
   value?: string;
   withCopyButton?: boolean;
   withOpenMapButton?: boolean;
+  isLink?: boolean;
```
resir014, line 32:
```suggestion
  hasOverflowScroll?: boolean;
```

Please use a more general purpose prop name.

resir014 (changes requested):
Some changes necessary, please take a look @abui-am.

components/contact-details.tsx
```diff
@@ -38,9 +39,16 @@ const DescriptionItem = (props: DescriptionItemProps) => {
     <div className="py-4 px-4 sm:py-5 sm:grid sm:grid-cols-3 sm:gap-4">
       <dt className="text-sm font-medium text-gray-500">{props.label}</dt>
       <dd className="mt-1 flex text-sm text-gray-900 sm:mt-0 sm:col-span-2">
-        <span className="flex-grow">
-          {htmr(value, { transform: htmrTransform })}
-        </span>
+        {props.isLink && value.length > 0 ? (
+          <span className="overflow-y-scroll flex-grow whitespace-nowrap copybox border-b border-brand-500 mr-2">
```
resir014, line 43: Use `overflow-y-auto` instead of `overflow-y-scroll`. We want to hide the scrollbar when it's not needed for a better user experience.

resir014 (changes requested):

resir014:
On second thought, we can also fix this without adding any more unnecessary props. Tailwind has a [`truncate` class](https://tailwindcss.com/docs/text-overflow#truncate) for us to easily truncate any text.

Which means we only have to do this. All the links are still preserved:

```tsx
<span className="flex-grow truncate">
  {htmr(value, { transform: htmrTransform })}
</span>
```

Which solution do you prefer, @abui-am?

abui-am:
> On second thought, we can also fix this without adding any more unnecessary props + CSS classes. Tailwind has a [`truncate` class](https://tailwindcss.com/docs/text-overflow#truncate) for us to easily truncate any text.
> 
> Which means we only have to do this. All the links are still preserved:
> 
> ```tsx
> <span className="flex-grow truncate">
>   {htmr(value, { transform: htmrTransform })}
> </span>
> ```
> 
> Which solution do you prefer, @abui-am?

Thanks for the suggestion, but I feel the truncated link just didn't look good for our user experience?
How do you think about it?

resir014:
> Thanks for the suggestion, but I feel the truncated link just didn't look good for our user experience? How do you think about it?

Paging @baraeb92 for design advice. Do you prefer scroll overflow or link truncation?

baraeb92:
Quick question, can we just streat this in the same way as alamat? 

Ideally most link that is included here shouldn't be too long, most of the time it should only as long as alamat, so treating it in the same way as alamat might be better. 

So we have 3 option here visually speaking: 

1. We can enable scroll so user can scroll the text 

2. We can enable long text like alamat field

3. We can enable truncate text

So let's take a look what the user might be intended for this

1. User want to go to the link so they can open it on the browser

2. User want to copy the link so they can share it. 

I think that is the most common use case right? 

Given that we already have salin button, I think use case number 2 is already taken care of. 

And if user only want to navigate to that link, that it wouldn't matter wether we are showing full link or truncated link, as long as it is clickable. 

So my opinion is that visual option 2 and 3 would be sufficient @resir014

abui-am:
I have an idea,
can we just treat it like a link?

I think the user will 70% likely just click it rather than copying it.
It solves our problem :
1. The link is shortened
2. User can go to the link just by clicking it
3. User can just copy it too

It will be looked like this
image

resir014:
@abui-am I don't think that's possible since the database already returns raw HTML of the link, hence why we transform it via `htmr`.

Maybe @zainfathoni can shed some light into this.

zainfathoni:
> @abui-am I don't think that's possible since the database already returns raw HTML of the link, hence why we transform it via `htmr`.

Yes, if we force it to treat it this way, it potentially confuses the editors because the transformation behaviour from the Spreadsheet data into the displayed value in the website is too cryptic.

> Maybe @zainfathoni can shed some light into this. Otherwise we can go ahead with option 3, as I initially suggested (truncated link)

I second that, let's go with the option 3.

resir014:
Noted @zainfathoni. @abui-am, please go ahead with link truncation, as suggested here:

https://github.com/kawalcovid19/wargabantuwarga.com/pull/562#issuecomment-893135749

abui-am:
I made a change, please take a look

resir014 (approved):
All good now. Thanks!

resir014:
@all-contributors please add @abui-am for code

PR #589, opened 2021-08-08
fix(button): align address with button

suryo182:
- Remove margin top under address p tag and add margin top to the address container
- Add col gap between button under contact details

Closes 

## Current Tasks

- [ ] 
- [ ] 
- [ ]


resir014 (changes requested):
@suryo182 Please fill in the PR description according to the template before we could look into this. Thanks.

PR #592, opened 2021-08-08
feat: hides/shows header on scroll

bungandy:
# Hides/shows header on scrolling a page

### Objective
If user is scrolling down, header area turn into invisible format to increase the area, so user can have more 'room' or reducing distraction elements. And when user is scrolling up, the header / burger menu become visible.

I know the header area quite small as it has 64px height, but fixed header has the downside of taking up a significant portion of the smaller mobile screen. The same pattern can be also seen in mobile safari or any browser app.

### Purpose
<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

https://user-images.githubusercontent.com/290786/128641828-c515bf4f-e9d7-4200-8943-f8d0a09544d6.mov

</td>
<td>

https://user-images.githubusercontent.com/290786/128641748-f473e5e7-f9a9-4fd3-a091-df38f4566acb.mov

</td>
</tr>
</table>

___

### Current Tasks
- [x] install dep using `yarn`
- [x] check unused state `lib/scroll-throttled.tsx`
- [x] add tests `global-header.tsx`
- [x] tests `lib/scroll-throttled.tsx` (deleted)


components/ui/scroll-throttled.tsx
```diff
@@ -0,0 +1,32 @@
+import { useState, useEffect } from "react";
+import { throttle } from "lodash";
```
zainfathoni, line 2: Please don't use `lodash`, it's bloating up the dependency.
If you really need it, please make sure that you're tree-shaking it properly and only include the `throttle` code in the bundle.

However, based on our [principles](https://kcov.id/wbw-principles), do you think it's worth adding this `throttle` dependencies to the home page, @resir014? 😁

zainfathoni (changes requested):

components/ui/scroll-throttled.tsx
mazipan: The bad thing from `Lodash`, it's not 100% tree-shake-able, since the code it self will use internal helpers.
So if we use it but only on small set, I will suggest to not use it completely

resir014: Yes @bungandy, please use a lighter alternative for the throttle, or don't add it at all. As our principles state:

> We must carefully consider any additional client-side libraries that we include on the website. Unless its benefits outweigh the trade-off that we have to pay, we should avoid adding the functionality.

mazipan: I think from the use case of `throttle`, it's replace-able with Debounce.
https://github.com/kawalcovid19/wargabantuwarga.com/blob/main/package.json#L44 we already have `ts-debounce`, maybe we can use it

bungandy: Hi mas @zainfathoni and @resir014 understood mas and thanks for the inputs

mas @mazipan I have managed to use custom throttle I put to `~/lib/..` folder.
Or we can use throttle from [@martinstark/throttle-ts](https://www.npmjs.com/package/@martinstark/throttle-ts) alternatively, kindly please check on the next commit, thanks 🙏

resir014: Bundle stats for [@martinstark/throttle-ts](https://bundlephobia.com/package/@martinstark/throttle-ts@1.2.3):

**Bundle size**

Minified: **725 B**
Gzipped: **431 B**

**Download time**

Slow 3G: **1ms**
Emerging 4G: **60μs**

This looks good. As @mazipan mentioned, we already have `ts-debounce` if you want to use libraries we have installed. If you can't use that, then feel free to use `throttle-ts`. @Bungandy

bungandy: @resir014 using throttle is better than debounce for header interaction I think mas, because debounce will always wait until the page is completely stopped when the page is scrolled whereas throttle will be executed in a certain time interval while the page being scrolled.

<table>
<tr>
<th>Throttle</th>
<th>Debounce</th>
</tr>
<tr>
<td>

https://user-images.githubusercontent.com/290786/128655850-84c5a63b-69e9-4c2d-a3f0-a0213975fa77.mov

</td>
<td>

https://user-images.githubusercontent.com/290786/128655805-7b167507-0653-4990-89b7-8309ad9c69ab.mov

</td>
</tr>
</table>

[reference](https://www.telerik.com/blogs/debouncing-and-throttling-in-javascript)

package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@headlessui/react": "^1.3.0",
     "@heroicons/react": "1.0.2",
+    "@martinstark/throttle-ts": "^1.2.3",
```
zainfathoni, line 30: Please install the new dependencies using `yarn`, so that the `yarn.lock` file would be updated as well and the CI check can pass.

components/layout/global-header.tsx
```diff
@@ -1,16 +1,42 @@
...
+      const isScrolledDown = previousScrollTop < currentScrollTop;
+      const isMinimumScrolled = currentScrollTop > MINIMUM_SCROLL;
+
+      setShouldShowShadow(currentScrollTop > 2);
+
+      setTimeout(() => {
+        setShouldHideHeader(isScrolledDown && isMinimumScrolled);
+      }, TIMEOUT_DELAY);
+    },
+  );
```
zainfathoni, line 31: Please add more tests to cover these new cases that you're introducing, thanks!

components/ui/scroll-throttled.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const [handleDocumentScrollThrottled] = throttle(handleDocumentScroll, 150);
+
+  useEffect(() => {
+    window.addEventListener("scroll", handleDocumentScrollThrottled);
+
+    return () =>
+      window.removeEventListener("scroll", handleDocumentScrollThrottled);
+  }, []);
+}
```
zainfathoni, line 31: If your new tests for the `global-header.tsx` is comprehensive enough, all these lines should have been covered.

components/ui/scroll-throttled.tsx
```diff
@@ -0,0 +1,33 @@
+import { useState, useEffect } from "react";
+import { throttle } from "@martinstark/throttle-ts";
+// import { debounce } from "ts-debounce";
+
+function useDocumentScrollThrottled(callback: {
+  (callbackData: { previousScrollTop: Number; currentScrollTop: Number }): void;
+  (arg0: { previousScrollTop: number; currentScrollTop: number }): void;
+}) {
+  const [, setScrollPosition] = useState(0);
```
zainfathoni, line 9: What's the point of setting the state here if you're not using the value at all?

zainfathoni (changes requested):

components/layout/global-header.tsx
bungandy: @zainfathoni do you have any references where I can start making a test for this? thank you

package.json
bungandy: @zainfathoni ok got it, thanks for the advice

components/ui/scroll-throttled.tsx
bungandy: @zainfathoni I've deleted this file and merged the function to `global-header.tsx` please check it again mas

components/ui/scroll-throttled.tsx
bungandy: @zainfathoni I've deleted this file and merged the function to `global-header.tsx` please check it again mas

components/layout/global-header.tsx
```diff
@@ -1,47 +1,55 @@
...
+const MINIMUM_SCROLL = 200; // 64 height header
+const TIMEOUT_DELAY = 150;
 
 export function GlobalHeader() {
   const popoverButtonRef = useRef<HTMLButtonElement>(null);
 
   const [shouldHideHeader, setShouldHideHeader] = useState(false);
-  const [shouldShowShadow, setShouldShowShadow] = useState(false);
+  const [, setScrollPosition] = useState(0);
+  let previousScrollTop = 0;
```
mazipan, line 19: You can try `useRef` to save the previous render value.

components/layout/global-header.tsx
```diff
@@ -1,47 +1,55 @@
...
+    const { scrollTop: currentScrollTop } = document.documentElement;
+    const isScrolledDown = previousScrollTop < currentScrollTop;
+    const isMinimumScrolled = currentScrollTop > MINIMUM_SCROLL;
 
-  useDocumentScrollThrottled(
-    (callbackData: { previousScrollTop: Number; currentScrollTop: Number }) => {
-      const { previousScrollTop, currentScrollTop } = callbackData;
-      const isScrolledDown = previousScrollTop < currentScrollTop;
-      const isMinimumScrolled = currentScrollTop > MINIMUM_SCROLL;
+    setScrollPosition((previousPosition) => {
```
mazipan, line 26: We can not use same name for arguments and upper scope variables, it cause a confusing

components/layout/global-header.tsx
mazipan: I don't think we need `setScrollPosition` in this case, just change to `useRef`

PR #602, opened 2021-08-09
docs: update README to add more requirements

resir014:
## Description

- Explicitly state that we use yarn, not npm.
- Explicitly state that English language is required.


mazipan (approved):
Ship it!

PR #659, opened 2021-08-12
feat: handle external url icon

### LINKED ISSUE #654: Capability to open new tab on Homepage icon
## Description

Based on issue #321 and PR #643, we add new icon 'kontribusi" which should be redirected to external URL.

The default behaviour is we need to open it on the new tab with proper `rel` attribute value.

## Tasks

- [ ] Add new `external` field on the icon list
- [ ] Handle properly for icons w `external: true`
- [ ] Open in new tab with `nofollow noreferrer noopener`

asaadam:
Closes #654

## Description

handle external URL to open it on the new tab with proper rel attribute value

## Current Tasks

- [X]  Add new external field on the icon list
- [X]  Handle properly for icons w external: true
- [X]  Open in new tab with nofollow noreferrer noopener

### DIFF
--- a/components/home/homepage-menu.tsx
+++ b/components/home/homepage-menu.tsx
@@ -20,7 +20,11 @@ export function HomePageMenu() {
             })}
           </div>
           <Link href={item.href}>
-            <a className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover">
+            <a
+              className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover"
+              rel={item.external ? "noopener noreferrer nofollow" : undefined}
+              target={item.external ? "_blank" : undefined}
+            >
               {item.name}
             </a>
           </Link>
--- a/components/home/homepage-menu.tsx
+++ b/components/home/homepage-menu.tsx
@@ -19,15 +19,22 @@ export function HomePageMenu() {
               className: "h-6 w-6 text-brand-500",
             })}
           </div>
-          <Link href={item.href}>
+          {item.external ? (
             <a
-              className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover"
-              rel={item.external ? "noopener noreferrer nofollow" : undefined}
-              target={item.external ? "_blank" : undefined}
+              className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover cursor-pointer"
+              href={item.href as string}
+              rel="noopener noreferrer nofollow"
+              target="_blank"
             >
               {item.name}
             </a>
-          </Link>
+          ) : (
+            <Link href={item.href}>
+              <a className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover">
+                {item.name}
+              </a>
+            </Link>
+          )}
         </li>
       ))}
     </ul>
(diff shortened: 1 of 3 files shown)


components/home/homepage-menu.tsx
```diff
@@ -20,7 +20,11 @@ export function HomePageMenu() {
             })}
           </div>
           <Link href={item.href}>
-            <a className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover">
+            <a
+              className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover"
+              rel={item.external ? "noopener noreferrer nofollow" : undefined}
+              target={item.external ? "_blank" : undefined}
+            >
```
mazipan, line 27: It can not like this, can not use next `Link` to open new tab.
You should create different tag.

asaadam: ahh yap 😭, I forget to re-check.

should we just <a> for implement it?

components/home/homepage-menu.tsx
```diff
@@ -19,11 +19,22 @@ export function HomePageMenu() {
...
+              href={item.href as string}
+              rel="noopener noreferrer nofollow"
+              target="_blank"
+            >
               {item.name}
             </a>
-          </Link>
+          ) : (
+            <Link href={item.href}>
+              <a className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover">
```
mazipan, line 33: Why we are not add `cursor-pointer` also in here. It should be identical

mazipan (changes requested):

components/home/homepage-menu.tsx
asaadam: somehow, it already has a cursor-pointer. 
I assume it's from the next link tag.

mazipan: Screen Shot 2021-08-12 at 20 56 09

asaadam: but also confused. 
why it has a cursor-pointer. 
image
I check here, but there's no CSS pointer 🧐

components/home/homepage-menu.tsx
```diff
@@ -19,11 +19,22 @@ export function HomePageMenu() {
               className: "h-6 w-6 text-brand-500",
             })}
           </div>
-          <Link href={item.href}>
-            <a className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover">
+          {item.external ? (
+            <a
+              className="text-gray-900 font-semibold text-xs mt-3 helper-link-cover cursor-pointer"
```
mazipan, line 24: I suggest to just remove the `cursor-pointer`

asaadam: ah I see, 
as long as tag `<a>` have attribute `href` it will already has cursor CSS

image

mazipan (approved):

components/home/homepage-menu.tsx
resir014: `helper-link-cover` is used to cover the entire list item with the link. That's the intended behaviour, since wrapping only the link text with `<a>` is better in terms of a11y.

PR #661, opened 2021-08-12
fix: display filtered `kebutuhan` contacts count on province-list

### LINKED ISSUE #581: Total count showing a wrong number when query param kebutuhan exist
## Description

- Open: Homepage > Klik "Donor Plasma"
  -  Or, just visit: https://www.wargabantuwarga.com/provinces?kebutuhan=Donor%20plasma directly
- "DKI Jakarta" show **459 Entry**
- But when you click it, it only have **4** data for "Donor Plasma"

Screenshot_20210807-185421_Chrome

## Expected Behaviour

When have query param `? kebutuhan=`, we expect user will only search for certain data in the province list, even they can change the selection on the province detail.

So my expectation also it will show the count of spesific "kebutuhan"

## Additional Information

We can work the quick fix for this, but I still prefer if we work on #486 first to make it easier for us differentiate the page.

nkristoporus:
Closes #581 

## Description

Fixed the issue of Province List displaying the total count of all contacts instead of the contacts of a specific `kebutuhan`. I added new fields in `provinces` props that count only contacts of a specific `kebutuhan`.

Screenshot 2021-08-12 at 8 52 40 PM

## Current Tasks

- [x] Fix Total count showing a wrong number when query param `kebutuhan` exist

### DIFF
--- a/__tests__/pages/provinces/index.test.tsx
+++ b/__tests__/pages/provinces/index.test.tsx
@@ -109,6 +109,15 @@ describe("getStaticProps", () => {
       name,
       slug,
       count: data.length,
+      ambulansCount: data.filter((contact) => contact.kebutuhan === "Ambulans")
+        .length,
+      rsCount: data.filter((contact) => contact.kebutuhan === "Rumah sakit")
+        .length,
+      donorPlasmaCount: data.filter(
+        (contact) => contact.kebutuhan === "Donor plasma",
+      ).length,
+      oksigenCount: data.filter((contact) => contact.kebutuhan === "Oksigen")
+        .length,
     }));
     expect(getStaticProps({})).toEqual({
       props: {
--- a/__tests__/pages/provinces/index.test.tsx
+++ b/__tests__/pages/provinces/index.test.tsx
@@ -58,44 +58,25 @@ describe("ProvincesPage", () => {
     ).toBeVisible();
   });
 
-  it("renders the provinces list kebutuhan entry count correctly", async () => {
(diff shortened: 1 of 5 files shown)


mazipan:
Can you check the code coverage, since it's decreased

__tests__/pages/provinces/index.test.tsx
```diff
@@ -57,6 +58,45 @@ describe("ProvincesPage", () => {
...
+      provinceListItem.rsCount,
+    );
+    await checkRenderKebutuhanEntryCount(
+      "Oksigen",
+      provinceListItem.oksigenCount,
+    );
+    await checkRenderKebutuhanEntryCount(
+      "Donor%20plasma",
+      provinceListItem.donorPlasmaCount,
+    );
```
zainfathoni, line 97: Instead of calling a custom function within the test like this, you can simplify the test and make the failures more obvious to the developers by using [Jest's Each](https://jestjs.io/docs/api#testeachtablename-fn-timeout) functionality. You can take a look at some other tests in this project that implements `.each()` as examples.

Please refactor this test to use `.each()` properly, thanks, @nkristoporus! 🙏

zainfathoni (changes requested):
Overall, the implementation looks good to me. 💯 

However, I have a minor issue with the test, please address it.
Thanks, @nkristoporus! 🙏

zainfathoni (approved):
LGTM 💯 

Thanks for your contribution, @nkristoporus! 🙏

zainfathoni:
@all-contributors please add @nkristoporus for code.

PR #636, opened 2021-08-11
feat: education page

iamyuu:
Closes #327 

## Description

Create an education page to replace `Laman Edukasi COVID-19` on the homepage

## Current Tasks

- [x] Test
- [x] Education page
- [x] Use real data 
- [x] config CMS for education data (#652)

PR #667, opened 2021-08-12
fix: remove non-critical image from initial load

mazipan:
## Description

Our homepage load many non-critical image on initial load. This PR try to solve that problem. Only load above the fold images.

## Current Tasks

- [x] Sembako CTA
- [x] Telemedicine CTA
- [x] Videos on About Page

## Before (7 images)

Screen Shot 2021-08-12 at 22 14 16

## After (3 images)

Screen Shot 2021-08-12 at 22 16 56


mazipan:
Blocked by #636

resir014 (approved):

PR #699, opened 2021-08-13
ci: increase number of run

mazipan:
## Description

Increase the number of run to 5, to reduce the instability CI

## Current Tasks

- [x] CI for PR
- [x] CI for Prod


zainfathoni (approved):
What an interesting approach. 🚀 

Thanks, Mas @mazipan! 🙏

PR #707, opened 2021-08-13
fix: revert whatsapp CTA button text color to white

resir014:
## Description

As discussed [here](https://kawalcovid19.slack.com/archives/C0279E2Q99Q/p1628866658042900?thread_ts=1628819593.022900&cid=C0279E2Q99Q).

Reverted the text colour for the WhatsApp button back to white, since the [WhatsApp branding guidelines](https://whatsappbrand.com/) require the logo to be white if the background is green.


mazipan (approved):

components/home/homepage-whatsapp-cta.tsx
```diff
@@ -9,7 +9,7 @@ export function HomePageWhatsAppCTA() {
       <h2 className="font-semibold">Masih belum dapat info yang kamu cari?</h2>
       <PrimaryAnchorButton
         block
-        className="text-gray-800 bg-[#25d366] focus:ring-[#25d366]"
+        className="text-white bg-[#25d366] focus:ring-[#25d366]"
```
zainfathoni, line 12: To simulate how people with vision deficiencies see our website, we can use this Google Chrome's feature, as described in [this article](https://www.theverge.com/2020/3/11/21174735/google-chrome-dev-tools-new-color-blind-friendly) or [this tweet](https://twitter.com/mathias/status/1237393102635012101).

Here's an illustration of how people with vision deficiencies see our website before and after this Pull Request:

https://user-images.githubusercontent.com/6315466/129425979-707e3e10-7588-46a3-b67b-6f6c6ac2e307.mov

What do you think, Mas @mazipan, Mbak @ekafyi, Mas @isatrio?
FYI, @kawalcovid19/wbw.

zainfathoni (commented):
Let's try to reconsider this change carefully.

components/home/homepage-whatsapp-cta.tsx
mazipan: Basically we are not using WhatsApp API.
I suggest to not using WhatsApp Logo in the button.

JusJust use heroicons

Screenshot_20210814-073555_Chrome

zainfathoni: This project might not be using any WhatsApp API, but I believe the Hotline project (our WhatsApp business integration with Qiscus and Kata.ai) uses it. Therefore, I think it's still relevant to use the official WhatsApp logo in that button. 😅

resir014:
@zainfathoni @mazipan From the [WhatsApp brand website](https://whatsappbrand.com/) we can use this colour instead:

image

So, should we remove the icon and replace all buttons on the website related to WhatsApp to the darker green as mentioned above?

zainfathoni:
If we want to keep the original WhatsApp color because it's more familiar for our users, I think we can use a Ghost button instead and use the WhatsApp green logo.

As illustrated in [this WhatsApp brand guide](https://whatsappbrand.com):

Screenshot 2021-08-14 at 13 09 40

resir014:
@zainfathoni Like this, then?

image

image

resir014:
Currently I'm also trying this combination, which IMO works better:

image

ekafyi:
> Currently I'm also trying this combination, which IMO works better:
> 
> image

I like this. Good contrast, visually still "whatsapp"-esque (for quick recognition UX-wise).

zainfathoni:
Yes, I think the last one works better, let's use that approach, then.

resir014:
alright leggooooo 🚀⛴️

zainfathoni (approved):
LGTM 💯 

Thanks, @resir014! 🙏

Issue #717, opened 2021-08-14
Fix warnings in `Tentang Kami` page test


zainfathoni:
## Description

Run this test

```shell
yarn test tentang-kami
```

And you'll see a lot of warnings due to the absence of `width` property in the `next/image` usages.

<img width="600" alt="Screenshot 2021-08-14 at 09 44 21" src="https://user-images.githubusercontent.com/6315466/129430803-81ebb704-856c-443a-92ca-6e5b65745c58.png">

## Expected Behaviour

No warnings should be shown

zainfathoni:
Probably the simplest way of fixing this issue is by adding a proper `width` props to the Image component.
What do you think, @resir014 & @ekafyi?

ekafyi:
> Probably the simplest way of fixing this issue is by adding proper `width` props to the Image component.
> What do you think, @resir014 & @ekafyi?

Yes, Next's Image component is supposed to have a width prop, like in Home page banners.

resir014:
Closed in #734.

PR #734, opened 2021-08-15
fix: remove warning tentang kami on test

rubiagatra:
Closes 
#717  Fix warnings in Tentang Kami page test

## Description

I think there is a paradox between ImageLoaderProps and Image when the layout fill

- You need to use width on ImageLoaderProps https://github.com/vercel/next.js/blob/daadfd347f1f66a725c7f85cf5dbf284010ec8a1/packages/next/client/image.tsx#L25
```
Image with src "v1628780633/128877540-e3a9b168-4fab-489c-8001-344347e3fb84_kce5oc.jpg" has a "loader" property that does not implement width. Please implement it or use the "unoptimized" property instead.
 Read more: https://nextjs.org/docs/messages/next-image-missing-loader-width
```

-  If you use "fill" layout you cannot use height || width  https://github.com/vercel/next.js/blob/canary/packages/next/client/image.tsx#L378
```
 Image with src "v1628562551/128724600-f7c60a95-9ab9-4761-bcbb-a25fa3596a1a_npxy45.png" and "layout='fill'"
 has unused properties assigned. Please remove "width" and "height".
  ```

## Current Tasks

- [x] Figuring out what happened and remove the warning
- [ ] Let's have some discussion to find the best way fix this


mazipan:
It's okay, I will try to re-upload all logos using a proper dimension.
So we can set the width.
Previously, I just force it to 200px since some images are too big.

resir014 (approved):

PR #737, opened 2021-08-15
fix: add CSP header for analytics in the correct paths

resir014:
## Description

...I really shouldn't make pull requests the first thing in the morning.


zainfathoni (approved):
😆

PR #738, opened 2021-08-15
fix: tweak logos section in the about us page

resir014:
## Description

Balanced out the logos section in the About Us page, since it's a bit all over the place right now with the `space` helper.

**Before:**

image

**After**

image


zainfathoni (approved):
LGTM 💯 

Thanks, @resir014!

PR #765, opened 2021-08-18
fix: replace rounded-md with rounded-t-md

Namchee:
Closes #764 

## Description

This pull request fixes unnecessary image rounding by replacing `rounded-md` with `rounded-t-md` in `DonationCard` component.

PR #766, opened 2021-08-18
fix: unnecessary border radius

hikhtiarta27:
This pull request fixes unnecessary border radius by replacing rounded-md with rounded-t-md in img tag from DonationCard component.


mazipan:
Duplicate with #765

PR #789, opened 2021-09-01
fix: minor fixes to homepage widgets

resir014:
### One.

Fixes the CTA button for the counseling widget to have `block` variant.

### Two.

Adds extra bottom padding on the WhatsApp widget to make space for the scroll to top button.

image


mazipan (approved):
Looks good

